PR #46, opened 2020-03-12
Add builtin rsqrt function

AndreiPotapov:
We have fast intrinsics for 1/sqrt and it's estimate version.

### DIFF
--- a/include/daScript/simulate/sim_policy.h
+++ b/include/daScript/simulate/sim_policy.h
@@ -147,16 +147,18 @@ namespace  das {
     };
 
     struct SimPolicy_MathFloat {
-        static __forceinline float Abs   ( float a, Context & )          { return v_extract_x(v_abs(v_splats(a))); }
-        static __forceinline float Floor ( float a, Context & )          { return v_extract_x(v_floor(v_splats(a))); }
-        static __forceinline float Ceil  ( float a, Context & )          { return v_extract_x(v_ceil(v_splats(a))); }
-        static __forceinline float Sqrt  ( float a, Context & )          { return v_extract_x(v_sqrt_x(v_splats(a))); }
-        static __forceinline float Min   ( float a, float b, Context & ) { return a < b ? a : b; }
-        static __forceinline float Max   ( float a, float b, Context & ) { return a > b ? a : b; }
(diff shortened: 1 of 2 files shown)


AntonYudintsev:
It seems rsqrt is better name, than invsqrt (HLSL, GLSL)

include/daScript/simulate/sim_policy.h
```diff
@@ -147,16 +147,18 @@ namespace  das {
...
-        static __forceinline float Floor ( float a, Context & )          { return v_extract_x(v_floor(v_splats(a))); }
-        static __forceinline float Ceil  ( float a, Context & )          { return v_extract_x(v_ceil(v_splats(a))); }
-        static __forceinline float Sqrt  ( float a, Context & )          { return v_extract_x(v_sqrt_x(v_splats(a))); }
-        static __forceinline float Min   ( float a, float b, Context & ) { return a < b ? a : b; }
-        static __forceinline float Max   ( float a, float b, Context & ) { return a > b ? a : b; }
-        static __forceinline float Sat   ( float a, Context & )          { return a < 0 ? 0 : (a > 1 ? 1 : a); }
-        static __forceinline float Mad   ( float a, float b, float c, Context & ) { return a*b + c; }
-        static __forceinline float Lerp  ( float a, float b, float t, Context & ) { return (b-a)*t +a; }
-        static __forceinline float Clamp  ( float t, float a, float b, Context & ){ return t>a ? (t<b ? t : b) : a; }
+        static __forceinline float Abs      ( float a, Context & )          { return v_extract_x(v_abs(v_splats(a))); }
```
AndreiPotapov: btw we don't need splats there

AntonYudintsev: well, we do

AntonYudintsev: it is float, and v_abs takes vec4f

AndreiPotapov: than we extracting only one component. nan's is not problem for simd instructions.

AntonYudintsev: yep
so on x86 (32 bit) it is different regs, and so the only way to do it

PR #63, opened 2020-06-24
crash fixes

mtsart:
- recursive deletion of HeapChunk list leads to stack overflow
- passing a non-ASCIIZ string to allocateString() leads to buffer overflow in internMap.find() which ignores the length


AntonYudintsev:
A bit uncertain about buffer overflow commit.

PR #148, opened 2021-07-16
fixed invalid position of error mark in error message

imp5imp5:


src/simulate/runtime_string.cpp
```diff
@@ -232,6 +232,7 @@ namespace das
             }
         }
         if ( row!=ROW ) return "";
+        auto beginOnLine = it;
```
AndreiPotapov, line 235: Of

imp5imp5:
fixed in https://github.com/GaijinEntertainment/daScript/pull/150

PR #253, opened 2021-12-11
Safier to safer

Falling10fruit:
Eh I just noticed something wrong with the webpage. Is this a accident or was this intended?


Falling10fruit:
Hmm, that's weird... It still says Safier instead of Safer
image

borisbat:
It does not update automatically. It involves pressing button. Should be good to go now.

Falling10fruit:
Thanks

PR #262, opened 2021-12-24
fix das tests

wurunduk:
commit https://github.com/GaijinEntertainment/daScript/commit/e7f62180c37b4cf4eefc1d46aa01d8427165121d
broke unit tests due to default float number format change

This commit should revert the tests to the state they were before the initial format change
(https://github.com/GaijinEntertainment/daScript/commit/31f6941b5d2c3cf4bdced1fc5fb14ace1ec70c92)


borisbat:
I've modified default flags for the string builder, so tests are no longer broken.

PR #268, opened 2022-02-14
Documented get_clock api

reOiL:
Added documentation for get_clock.
Taked from: https://www.cplusplus.com/reference/ctime/time/


profelis:
it's online, Thanks! https://dascript.org/doc/stdlib/builtin.html?highlight=get_clock#get_clock

PR #408, opened 2023-02-19
speedup ci builds

profelis:
restore debug configuration for windows ci builds


profelis:
https://github.com/GaijinEntertainment/daScript/pull/409

PR #466, opened 2023-03-06
static assert: try to return type larger than vec4f

reOiL:
Now we assert if binding try to return by value
and cast not exist

Now error now tells you what to do to fix it


borisbat:
comment is incorrect. has nothing to do with size of the structure. needs cast<>, or better yet cast_res<>.
test for the static_if is incorrect as well.

reOiL:
> comment is incorrect. has nothing to do with size of the structure. needs cast<>, or better yet cast_res<>. test for the static_if is incorrect as well.

Fixed

PR #625, opened 2023-06-07
Fix compile time warning for POD test.

rburkholder:
This is a warning generated when building with `-std=gnu++20`
[build] warning: ‘template<class _Tp> struct std::is_pod’ is deprecated: use is_standard_layout && is_trivial instead [-Wdeprecated-declarations]
[build]   246 |             return is_pod<OT>::value;
[build]       |                    ^~~~~~


borisbat:
it would need #if __cplusplus >= 202002L right?

PR #639, opened 2023-06-22
Add is_digit method

otakubeam:


otakubeam:
There is _daslib/strings_ `is_number` function

PR #671, opened 2023-07-08
Add PEG module documentation (initial version)

otakubeam:


otakubeam:
I will create new branches for commits now

PR #826, opened 2023-10-21
fix typos (in words "suppress", "beginning" etc.)

imp5imp5:


include/dag_noise/dag_uint_noise.h
borisbat, line 1: this is part of the dagor, btw

PR #835, opened 2023-10-29
Haiku port

jacereda:
Compiles with clang17 and gcc13. I switched some ``__APPLE__`` checks that are really clang-specific (maybe should check clang version instead).


include/daScript/simulate/aot_builtin_jobque.h
```diff
@@ -102,7 +102,7 @@ namespace das {
             }
             lock_guard<mutex> guard2(that->mCompleteMutex);
             for ( auto & f : pipe ) {
-                that->pipe.emplace_back(move(f));
+                that->pipe.emplace_back(std::move(f));
```
profelis, line 105: just move(f) please. All std related functions should be visible from das namespace.
We use eastl instead of default std and all this std:: will break our build.
Same problem in some next changes.

jacereda: Oh, OK. I added that to silence a warning, should I just add the relevant `-W-no-xxx`?

PR #836, opened 2023-10-31
add intern override with len

Bais:
It need for serialization with not null terminated strings

### DIFF
--- a/src/simulate/simulate.cpp
+++ b/src/simulate/simulate.cpp
@@ -1195,9 +1195,14 @@ namespace das
         });
     }
 
-    char * Context::intern(const char * str) {
+    char * Context::intern( const char * str ) {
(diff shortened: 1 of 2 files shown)


src/simulate/simulate.cpp
```diff
@@ -1195,9 +1195,14 @@ namespace das
         });
     }
 
-    char * Context::intern(const char * str) {
+    char * Context::intern( const char * str ) {
         if ( !str ) return nullptr;
         uint32_t len = uint32_t(strlen(str));
+        intern(str, len);
```
profelis, line 1201: return intern(str, len);

Bais: https://github.com/GaijinEntertainment/daScript/pull/837

PR #839, opened 2023-11-03
support partial serialization

otakubeam:


include/daScript/ast/ast_serializer.h
```diff
@@ -25,6 +25,9 @@ namespace das {
         bool                writing = false;
         size_t              readOffset = 0;
         vector<uint8_t>     buffer;
+        vector<uint8_t>     metadata;
+        function<int64_t ( const char * )> dagor_get_mtime;
```
borisbat, line 29: thats an interesting naming

src/ast/ast_parse.cpp
```diff
@@ -461,6 +506,190 @@ namespace das {
...
+            addNewModules(libGroup, program);
+        }
+        auto res = parseDaScript(fileName, access, logs, libGroup, exportAll, false, policies);
+        policies.threadlock_context |= res->options.getBoolOption("threadlock_context",false);
+        if ( !res->failed() ) {
+            if ( res->options.getBoolOption("log_symbol_use") ) {
+                res->markSymbolUse(false, false, false, nullptr, &logs);
+            }
+        }
+        if ( policies.aot_module && (res->promoteToBuiltin || res->thisModule->isModule || exportAll) ) {
```
borisbat, line 643: thats a lot of copy-paste from the function bellow. is it a good idea to split so much?

otakubeam: The idea was to split features from refactoring. Now that I think about it, makes no sense. Fixed

PR #841, opened 2023-11-06
Prepare for packaging

jacereda:


jacereda:
I haven't tested the Windows/Mac builds. Not a cmake expert, but I think this should prepare the project for unix packaging. 

I have removed the stripping option, it yields warnings in some compilers because it's a link option.  IMHO this is better handled by the `--strip` option of `--install`.

I also fixed some unused variable warnings.

PR #878, opened 2023-12-18
Fix thread local debug agent install

reOiL:
Before we don't call onInstall for thread local debug agent,
but call for other agents


reOiL:
Not finished

PR #972, opened 2024-02-15
Inherited 'operator dot' is not callable when declared in separate file

frodiwe:
Hi!

I run upon a case where you are not able to call inherited `operator .` function because of `field not found in (30503)`.
I have added test case for the issue in this PR.

The case seems to be like this:

File_A: declare base `class (A)` with `operator .`
File_B: declare derived `class B` from `class A`
File_C: `require File_B`, declare variable of type `class B` and call  `operator .` (it should be there because `class B` is derived from `class A`)

It is fixable by requiring `File_A` in `File_C` (along with `File_B`) but I suppose it is not how it was intended to work.

### DIFF
--- a/tests/language/operators.das
+++ b/tests/language/operators.das
@@ -2,6 +2,8 @@ require dastest/testing_boost public
 
 require daslib/strings_boost
 
+require operators_derived
+
 struct Foo
     a : int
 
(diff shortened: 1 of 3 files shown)


borisbat:
"require operators_parent public" is important (vs "require operators_parent"). it makes stuff in operators_parent visibile. I've merged tests and fixed the require too make sure it works as intended.

PR #1166, opened 2024-07-10
Add split_by_chars_with_quotes function to ignore part of string for …

wurunduk:
…splitting

split_by_chars version which stops splitting if we are inside
some other specified delimiter.

Useful when we want to split a string like
`some.cmd "param with space" 123`
into
[[ some.cmd; "param with space"; 123]]

for example above call looks like `input |> split_by_chars_with_quotes(" ", "\"")`

### DIFF
--- a/daslib/strings_boost.das
+++ b/daslib/strings_boost.das
@@ -72,6 +72,15 @@ def split_by_chars ( text, delim : string implicit ) : array<string>
         res := arr
     return <- res
 
+def split_by_chars_with_quotes ( text, delim, quotes : string implicit; blk : block< (arg:array<string>#) > )
+    builtin_string_split_by_char_with_quotes ( text, delim, quotes, blk )
+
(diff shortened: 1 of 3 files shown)


borisbat:
given the complexity of the problem, please add tests, similar to that in split.das

wurunduk:
> given the complexity of the problem, please add tests, similar to that in split.das

Dmitri suggested that this solution is problem specific enough that putting this in das itself does not make sense.
What do you think? I can add tests for this, but maybe he is right and this commit is not really needed

borisbat:
its a whole can of warms. like for instance, do we support escaped strings, and if so - what are the escaping rules. and all that.
there is peg/peg library for parsing complex grammars, but im not sure if this qualifies or not.
this is one of those borderline cases - where it feels like it can go either way, but just sheer amount of test cases necessary sways me in the direction of 'too specific'.
either way, if it is to go in, it needs support for escaping, bunch of tests, and all that.

PR #1228, opened 2024-08-02
Enum, bitfield & bools for typemacro_function

Mrkol:
These plus ints cover a whole bunch of use-cases of the typestate programming pattern. Now we wait for matching.


borisbat:
https://github.com/GaijinEntertainment/daScript/pull/1229

PR #1247, opened 2024-08-23
Fixed clang compilation issues (POSIX function names mostly).

e-dog:


borisbat:
Does not work on linux, wasm, or osx. Perhaps an ifdef on windows, or some such?

e-dog:
So, it's a clang + ms runtime thing... I'll try again with defines then.

PR #1268, opened 2024-09-12
Add framework for pretty benchmark results building and formatting

todace:
hasn't test it fully cause there is no daslang.exe to benchmark, and no mono
So, just convert data from html to json and reformat it


borisbat:
https://github.com/GaijinEntertainment/daScript/pull/1291

PR #1308, opened 2024-10-13
if and while report idented block

borisbat:


borisbat:
Ended up a bad idea. There are other errors which manifest the same way.

PR #1392, opened 2024-11-15
Rework dasException

reOiL:
There a problem with exception if we catch it outside of das::Context scope we can get broken memory for 'what' pointer due the fact we storage exception message inside das::Context and put only pointer to exception

Because of that i would like to copy exception message to dasException I decide to use runtime_error, because:
`It reports errors that are due to events beyond the scope of the program and cannot be easily predicted.`

### DIFF
--- a/include/daScript/simulate/simulate.h
+++ b/include/daScript/simulate/simulate.h
@@ -203,16 +203,12 @@ namespace das
     } }
 
 #if DAS_ENABLE_EXCEPTIONS
-    class dasException final : public std::exception {
+    class dasException final : public std::runtime_error {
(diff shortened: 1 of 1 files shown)


borisbat:
needs custom exception class, which would store it.
because "Собственно у нас на сони этот код-пасс крешится. В частности потому что std::string биранно несовместимый между libc & приложением. А у std::runtime_error деструктор со строкой находится в стандатной либе..."

reOiL:
Update: replace runtime_error exception with std::exception and std::string to avoid crash if libc is not binary compatible to application

PR #1413, opened 2024-12-03
Added parseCustomRequire() to process prerequisites added via reader 

e-dog:
…macros.


e-dog:
It's better to ignore whitespace changes when viewing ast_parse.cpp diff.

PR #1505, opened 2025-02-10
new structure init to accept aliases

borisbat:
```
variant JsValue {
    _array  : array<JsonValue?>;
    _string : string;
}
struct JsonValue {
    value : JsValue;
}

[export]
def main() {
    let v = "";
    let v1 = JsValue(_string=v);    // NOW OK
    debug(v1);
}
```


src/parser/ds_parser.ypp
```diff
@@ -3500,6 +3497,7 @@ use_initializer
 
 make_struct_decl
     :   BRABRAB [loc] type_declaration_no_options[mkt] make_struct_dim[msd] optional_block[blk] optional_trailing_delim_sqr_sqr {
+        TextPrinter tp;
```
aleksisch, line 3500: Shouldn't we delete it?

borisbat: I just did

PR #1518, opened 2025-02-20
Bug/fix chars in format

aleksisch:
This PR fixes bug in das_source_formatter and format remaining .das files

Originally bug was:
```
def foo() {
    let x = '{';
}
let y = '\t'
```
is formatted to
```
def foo() {
    let x = '{';
}
let y = ' \ t';
```
It caused by incorrect handling of characters and strings


examples/test/compilation_fail_tests/invalid_type_mix.das
```diff
@@ -2,7 +2,7 @@ expect 30101:3, 30108:1    // invalid_type
 
 def test(A : void&)     // 30101: can't be void reference
                         // 30101: function argument type can't be declared void
-    let b : int&&       // 30108: local reference needs to be initialized
```
aleksisch: Is && aroung type valid? Now it says that it's just incorrect

examples/test/compilation_fail_tests/test_require_comment.das
```diff
@@ -8,7 +8,7 @@ expect 30304
 some random text
 require _test2_require_comment
 
-* /
```
aleksisch: We can't parse comment with space between * and /. Was this test valid? Expected behavior was to ensure that we don't process code inside comments

imp5imp5:
I would suggest a shorter fix for formatter: https://github.com/GaijinEntertainment/daScript/pull/1519

aleksisch:
> I would suggest a shorter fix for formatter: #1519

Thanks. That's better

aleksisch:
There was another similar bug with nested comments
```
/*
/*
*/
*/
let x = 123;
```

PR #1540, opened 2025-02-28
enumeration and bitfield error reporting and visibility rules

borisbat:


src/ast/ast_infer_type.cpp
```diff
@@ -5440,21 +5440,24 @@ namespace das {
...
                 vector<Enumeration *> possibleEnums;
                 vector<TypeDecl *> possibleBitfields;
                 program->library.foreach([&](Module * mod) -> bool {
-                    if ( auto possibleEnum = mod->findEnum(enumName) ) {
-                        possibleEnums.push_back(possibleEnum.get());
-                    }
-                    if ( auto possibleBitfield = mod->findAlias(enumName) ) {
-                        if ( possibleBitfield->isBitfield() ) {
-                            possibleBitfields.push_back(possibleBitfield.get());
+                    if ( inWhichModule->isVisibleDirectly(mod) ) {
```
aleksisch, line 5453: Maybe invert if and make early return if invisible?

borisbat: same thing, no?

PR #1551, opened 2025-03-04
Standalone contexts

aleksisch:
This PR finalize work on `standalone contexts`


src/simulate/simulate.cpp
```diff
@@ -1413,7 +1413,7 @@ namespace das
         return runWithCatch([&](){
             for ( int j=0, js=totalFunctions; j!=js && !stopFlags; ++j ) {
                 auto & pf = functions[j];
-                if ( pf.debugInfo->flags & FuncInfo::flag_shutdown ) {
+                if ( pf.debugInfo && pf.debugInfo->flags & FuncInfo::flag_shutdown ) {
```
aleksisch, line 1416: Looks like it's not quite easy to generate debugInfo inside standalone context, maybe we'll add it later. 
Now we'll fail in destructor due to nullptr dereference.
Is it fine @borisbat?

borisbat: we already generate debug info in AOT. should be very doable

PR #1590, opened 2025-03-19
[WIP] expression hash tags for lookup

borisbat:


borisbat:
this somehow slows things down. im not going to debug why.

PR #1594, opened 2025-03-20
enable aot for ast_boost

aleksisch:
Fix few tiny bugs


include/daScript/simulate/aot.h
```diff
@@ -628,11 +628,8 @@ namespace das {
     }
 
     template <typename TT>
-    struct das_index<vector<TT>> : das_default_vector_index<vector<TT>, TT> {};
-
-    template <typename TT>
-    struct das_index<vector<TT> const> : das_default_vector_index<vector<TT>, TT> {};
-
+    struct das_index<TT, std::enable_if_t<std::is_base_of_v<std::vector<typename TT::value_type>, TT>>>
```
aleksisch, line 631: We need to allow struct-s such as `AnnotationList`, which inherited from `std::vector` to be used in `das_index`.

src/ast/ast_aot_cpp.cpp
```diff
@@ -1381,7 +1381,8 @@ namespace das {
                 ss << "das_auto_cast<" << describeCppType(var->type) << ">::cast(";
             }
             if ( expr->type->isString() ) {
-                ss << "(char *)(";
+                auto maybeRef = var->type->ref ? " &" : "";
```
aleksisch, line 1384: @boris can you verify, that this diff is correct?

include/daScript/simulate/aot.h
borisbat: you can specialize it in aot_builtin_ast.h or some such, no?

src/ast/ast_aot_cpp.cpp
borisbat: looks good

PR #1611, opened 2025-03-27
Add support for aot on almost whole files in stdlib (but not all func…

aleksisch:
…tions).

Now it fails to aot temp_strings.das due to outer block, which is not suppported in aot yet/

### DIFF
--- a/daslib/dap.das
+++ b/daslib/dap.das
@@ -50,23 +50,6 @@ struct Capabilities {
 
 
 
-
-// https://microsoft.github.io/debug-adapter-protocol/specification#Requests_SetBreakpoints
-struct SetDataBreakpointsArguments {
-    breakpoints : array<DataBreakpoint>
-}
-
-def SetDataBreakpointsArguments(data : JsonValue?) {
-    var res <- SetDataBreakpointsArguments()
-    var breakpoints = joj(data, "breakpoints")
-    if (breakpoints != null) {
-        for (it in breakpoints.value as _array) {
-            res.breakpoints |> emplace(DataBreakpoint(it))
-        }
-    }
-    return <- res
-}
-
 // https://microsoft.github.io/debug-adapter-protocol/specification#Types_DataBreakpoint
 struct DataBreakpoint {
     dataId : string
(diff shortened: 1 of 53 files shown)


daslib/dap.das
```diff
@@ -50,23 +49,6 @@ struct Capabilities {
 
 
 
-
```
aleksisch, line 53: Das allows to `declare after use`, `c++` not. For now I think we should just put everything in right order in das files

daslib/debug.das
```diff
@@ -176,7 +175,7 @@ class DAWalker : DapiDataWalker {
 
     varsStack : array<DAVariable?>
     @do_not_delete frame : DAStackFrame?
-    visited : array<tuple<ps : void?; hash : uint64>>
+    visited : array<tuple<ps : void const?; hash : uint64>>
```
aleksisch, line 178: @borisbat Is it bug in das? Later, on 226 line we call
```
    def override beforeStructure(ps : void const?; si : StructInfo) : void {
        visited |> emplace((ps, si.hash))
    }
``` 
I guess it should not be possible, due to const-cast?

daslib/debug.das
```diff
@@ -2248,7 +2247,7 @@ class private DAServer : Server {
             "text" => empty(text) ? JVNull() : JV(text),
             "threadId" => JV(double(ctx.id)),
             "allThreadsStopped" => JV(false),
-            "hitBreakpointIds" => brId == -1ul ? null : JV([ JV(brId)])
+            "hitBreakpointIds" => brId == -1ul ? JVNull() : JV([ JV(brId)])
```
aleksisch, line 2250: Aot resolves it to `void*` instead of `JsonValue *`. I believe it's incorrect?

daslib/temp_strings.das
```diff
@@ -17,7 +16,7 @@ def public build_temp_string(bldr : block<(var writer : StringBuilderWriter) : v
     unsafe {
         cb |> invoke(reinterpret<string#> str)
         if (!is_intern_strings()) {
-            delete_string(reinterpret<string> str)
+            delete_string(reinterpret<string&> str)
```
aleksisch, line 19: `delete_string` assigns str to nullptr. Was previous code correct? We were passing `str` and still got `nullptr` after the call

daslib/debug.das
borisbat: das trinary operator resolves different from C++, easier to use.

daslib/temp_strings.das
borisbat: i wonder how it worked in the interpreter

PR #1615, opened 2025-04-01
Remove cpp printer

aleksisch:


aleksisch:
This PR blocked by different ast_aot output on 32-bit and 64-bit platforms

PR #1635, opened 2025-04-11
Support mult line expressions

aleksisch:


utils/dasFormatter/ds_parser.ypp
```diff
@@ -2196,14 +2195,7 @@ expr_call
     ;
 
 expr_not_wrapped
```
aleksisch, line 2197: We should wrap only necessary expressions.
For example this code allowed in gen2: 
```
def main() {
    let x = new A(x=1,
    y=2);
}
```

PR #1641, opened 2025-04-21
Feature/ast aot cpp das

aleksisch:


daslib/ast_boost.das
```diff
@@ -1037,6 +1094,13 @@ def public isMakeLocal(expr : ExpressionPtr) {
         ||  (expr.__rtti == "ExprMakeTuple"))
 }
 
+def public isExprCallFunc(expr : ExpressionPtr) {
```
aleksisch, line 1097: In C++ we can call `rtti_isCallFunc` and it will check if expression is `CallFunc` or it's child, however it doesn't work in das. Can we do it in das?

PR #1664, opened 2025-05-05
Added semicolon support for formater and fix tmp file path

reOiL:


borisbat:
does not merge any more. i've fixed temp path in a similar fashion already :(

PR #1667, opened 2025-05-06
Added thread local wrapper

reOiL:
Why: because its easy replace without workarounds like overload all possible operators

Interface: DAS_THREAD_LOCAL(`<Type>`)


borisbat:
lets at least add a macro. those magic numbers are not very readable

borisbat:
also, we need 2nd define. to enable\disable all this machinery, and roll back to the regular threadlocal

reOiL:
> also, we need 2nd define. to enable\disable all this machinery, and roll back to the regular threadlocal

The idea is just a common interface for thread_local. Making a backport for thread_local X is not possible because we will have to redefine all potentially used operators/methods + some language constructs will not be possible, for example auto X = MyThreadLocal.

examples/profile/test_profile.cpp
```diff
@@ -441,10 +441,10 @@ struct QueryEsFunctionAnnotation : FunctionAnnotation {
...
 };
 
-DAS_THREAD_LOCAL(vector<float3>)          g_pos;
-DAS_THREAD_LOCAL(vector<float3>)          g_vel;
-DAS_THREAD_LOCAL(vector<float3 *>)        g_velBoxed;
-DAS_THREAD_LOCAL(vector<EsComponent>)     g_components;
+DAS_THREAD_LOCAL(vector<float3>, 0x01c67c0a)          g_pos;
+DAS_THREAD_LOCAL(vector<float3>, 0x72ace5c7)          g_vel;
+DAS_THREAD_LOCAL(vector<float3 *>, 0xf81c229b)        g_velBoxed;
+DAS_THREAD_LOCAL(vector<EsComponent>, 0xc19bf781)     g_components;
```
aleksisch, line 447: Let's add `consteval crc32` function?

PR #1679, opened 2025-05-19
[Do not merge] ]add repro

aleksisch:
```
make ./bin/daScriptTest
./bin/daScriptTest src/das/main.das -- --mode aot --input examples/test/unit_tests/add_property_ext_const.das --output test2.cpp
```
This command leads to memory corruprion


aleksisch:
It happened due to
```
            var l_clone <- unsafe(reinterpret<ExpressionPtr>(that.subexpr.get_ptr())); // todo: is there a better way
            that.arguments |> emplace(l_clone);
```

PR #1688, opened 2025-05-23
use smart_ptr_raw in aot

aleksisch:


include/daScript/ast/ast_typedecl.h
```diff
@@ -674,7 +674,8 @@ namespace das {
                            CpptSubstitureRef substituteRef = CpptSubstitureRef::no,
                            CpptSkipRef skipRef = CpptSkipRef::no,
                            CpptSkipConst skipConst = CpptSkipConst::no,
-                           CpptRedundantConst redundantConst = CpptRedundantConst::yes );
+                           CpptRedundantConst redundantConst = CpptRedundantConst::yes,
+                           bool chooseRawSmartPtr = false );
```
aleksisch, line 678: @borisbat What do you think? Is there a better way to handle special case with `das::vector` when we need to keep vector of smart pointers?

borisbat: i think its not always right. i.e. do we always use ManagedVectorAnnotation for das::vector? can we doulbe-check? other templates can be bound the same way

PR #1692, opened 2025-05-26
enable ast quote in aot

aleksisch:


daslib/ast_boost.das
```diff
@@ -834,6 +849,275 @@ def convert_to_expression(var value : auto& ==const; at : LineInfo) {
...
+        }
+        // a[i].at := e.at
+        a[i].basicType = e.basicType
+    }
+}
+
+def private get_vec_dst_type(var src_tp : TypeDeclPtr) {
+    let tstr = describe(src_tp);
+    if (tstr |> starts_with("$::das_string")) {
+        delete src_tp;
```
aleksisch, line 927: Can we avoid `delete src_tp` here? Something like `var inscope` in argument?

src/builtin/builtin.das
```diff
@@ -1116,7 +1116,7 @@ def clone(var a : table<string; auto(VT)>; b : table<string; VT> | #) {
 }
 
 [expect_any_vector(args), skip_lock_check]
-def clone(var args; var nargs : array<string>) {
+def clone(var args; var nargs : array<auto(TT)>) {
```
aleksisch, line 1119: I guess it should be removed. Now we need it to convert arrays of primitive types to vectors

daslib/ast_boost.das
aleksisch: Made a move with var inscope

src/builtin/builtin.das
aleksisch: Implemented for all required primitives

src/builtin/builtin.das
```diff
@@ -1125,7 +1125,7 @@ def clone(var args; var nargs : array<string>) {
     // delete nargs // we can make this unsafe_when_not_clone_array, or just let GC deal with it
 }
 
-[expect_any_vector(args), skip_lock_check, unsafe_when_not_clone_array]
+[expect_any_vector(args), skip_lock_check]
```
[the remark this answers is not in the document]
aleksisch: It was another bug, fixed

PR #1696, opened 2025-05-27
never rebuild daslang-stage0

aleksisch:


.github/workflows/build.yml
```diff
@@ -163,19 +163,23 @@ jobs:
               linux64)
                 cmake --no-warn-unused-cli -B./build -DCMAKE_BUILD_TYPE:STRING=${{ matrix.cmake_preset }} -G "${{ matrix.cmake_generator }}"
                 cd build
+                ninja daslang-stage0
```
aleksisch, line 166: For now we should compile daslang-stage0 manually

PR #1708, opened 2025-05-30
Add dasgen validation to CI

aleksisch:


src/ast/ast_handle.cpp
```diff
@@ -19,10 +19,9 @@ namespace das {
 
     uint64_t BasicStructureAnnotation::getOwnSemanticHash ( HashBuilder & hb, das_set<Structure *> & dep, das_set<Annotation *> & adep ) const {
         hb.updateString(getMangledName());
-        for ( auto & it : fields ) {
-            auto & sfield = it.second;
+        for ( auto & name : fieldsInOrder ) {
+            auto & sfield = fields.at(name);
             hb.updateString(sfield.name);
-            hb.update(sfield.offset);
```
aleksisch, line 25: Offset depends on platform. @borisbat I guess without it we're fine here?

src/simulate/simulate_fn_hash.cpp
```diff
@@ -119,14 +121,17 @@ namespace das {
...
+        debug_hash("\nresult <%s> = %" PRIx64 "\n", fun->result->getMangledName().c_str(), resT);
         for ( auto & arg : fun->arguments ) {
             uint64_t argT = arg->type->getSemanticHash();
             hashV.write(&argT, sizeof(argT));
-            debug_hash("arg %s <%s> = %llx\n", arg->type->getMangledName().c_str(), arg->name.c_str(), argT);
+            debug_hash("arg %s <%s> = %" PRIx64 "\n", arg->type->getMangledName().c_str(), arg->name.c_str(), argT);
         }
+
+        // Use mangled name instead of function body
+        const auto mnh = fun->getMangledNameHash();
```
aleksisch, line 132: I suggest to use mangledNameHash instead of hash from code. However it creates at least one visible change: if we have two identical functions with different names in current master we will have same hash for them. After this commit it will be different.

However if necessary we still able to support this feature, for example by introducing `map` from old hashes to new ones. If old hash for current function already presented it means we should choose value from `map`

.github/workflows/build.yml
```diff
@@ -185,7 +185,21 @@ jobs:
...
+      # Ignore changes which depends on stack top.
+      # Also ignore parser generated files.
+      run: |
+        set -eux
+        case "${{ matrix.target }}${{ matrix.architecture }}" in
+           windows32)
+            # windows32 is not ready yet
+            ;;
+           *)
+            git diff --ignore-all-space --text -I".*FunctionInfo.*" -I".*FuncInfo.*" -I".*das_stack_prologue.*" -I".*smart_ptr_raw<Function>.*" --exit-code -- ':!src/parser/ds2_lexer.cpp' ':!src/parser/ds_lexer.cpp' ':!src/parser/ds2_parser.output' ':!src/parser/ds_parser.output' ':!src/parser/lex2.yy.h' ':!src/parser/lex.yy.h' ':!utils/dasFormatter/ds_parser.cpp' || (echo "Generated code changed! To reproduce use `./bin/daslang dasgen/gen_bind.das`." && exit 1)
```
aleksisch, line 200: I suggest to add this check to CI to verify that all files remains the same.
Also it will check that `gen_bind` files are up-to-date

PR #1751, opened 2025-07-01
Support releases in CI

aleksisch:


.github/workflows/release.yml
```diff
@@ -0,0 +1,241 @@
...
+                linux64)
+                  cmake --no-warn-unused-cli -B./build -DCMAKE_BUILD_TYPE:STRING=${{ matrix.cmake_preset }} \
+                    -G "${{ matrix.cmake_generator }}" $ACTIVE_MODULES
+                  cd build
+                  ninja daslang
+                  ;;
+                *)
+                  export CXXFLAGS="-Wno-elaborated-enum-base" # glfw module fails due to mac-os includes
+
+                  CC=clang CXX=clang++ cmake --no-warn-unused-cli -B./build -DCMAKE_OSX_ARCHITECTURES="${{ matrix.target == 'darwin' && 'arm64' || 'x86_64' }}" \
```
aleksisch, line 197: @profelis Looks like we should do the same in `build.yml`? Idk why it compiles with `x86_64` flag

profelis: yep! You're right! Btw you can use build settings for what. See how {{ matrix.cmake_preset }} works for example.

PR #1755, opened 2025-07-02
fix most of CE in das files

aleksisch:


examples/contDev/main.das
```diff
@@ -1,103 +1,124 @@
...
+def cppTypeName(info: VarInfo) {
     var res = cppTypeAttr(info)
-    if res == ""
+    if (res == "") {
         res = info.isConst ? "const " : ""
-        if info.basicType == Type tHandle
-            res += string(info.annotation_or_name.name)
-        elif info.basicType == Type tStructure
+        if (info.basicType == Type.tHandle) {
+            res += string(unsafe(reinterpret<TypeInfo>(info)).annotation_or_name.name) // Why do we need cast?
```
aleksisch, line 89: @borisbat Is it correct? Why it doesn't compile without casts, VarInfo inherited from TypeInfo in bindingd

examples/test/misc/context_state_example.das
```diff
@@ -1,103 +1,131 @@
...
-    def override onCall(pp : Prologue; info : FuncInfo) : void
+    }
+    def override onCall(pp : Prologue; info : FuncInfo) : void {
         print("{info.name}\n")
-    def override onArgument(info : FuncInfo; index : int; vinfo : VarInfo; arg : float4) : void
+    }
+    def override onArgument(info : FuncInfo; index : int; vinfo : VarInfo; arg : float4) : void {
         print("\targ {vinfo.name} ")
-        print(sprint_data(arg, type_info(vinfo), print_flags humanReadable))
+        print(sprint_data(arg, type_info(vinfo), default<print_flags>))
```
aleksisch, line 21: Does default equivalent to `humanReadable`?

borisbat: nop. it should be print_flags.humanReadable i think

examples/contDev/main.das
borisbat: not sure. maybe for historical reasons. have u tried without?

examples/test/misc/context_state_example.das
aleksisch: `humanReadable` was deleted from this binded enum. `singleLine` instead? 
https://github.com/GaijinEntertainment/daScript/commit/bc8acb4cac53bb2ca3ef9346aa94b6e3cba363f4

PR #1994, opened 2025-12-26
No more cpp jit

aleksisch:


modules/dasLLVM/daslib/llvm_jit.das
```diff
@@ -905,20 +905,22 @@ class LlvmJitVisitor : AstVisitor {
...
+        var vsize = arg._type.flags.ref ?  typeinfo sizeof(type<void?>) : arg._type.sizeOf
+        var valign = arg._type.flags.ref ?  typeinfo alignof(type<void?>) : arg._type.alignOf
+
         if (arg.flags.aliasCMRES) {
             v_ptr = get_cmres_param()
             v_ptr = LLVMBuildPointerCast(g_builder, v_ptr, vptrtype, "variable_cmres_{arg.name}")
         } else {
             at_function_entry <| $() {
                 v_ptr = LLVMBuildAlloca(g_builder, vtype, "variable_{arg.name}")
+                LLVMBuildMemSet(g_builder, v_ptr, 0ul, uint64(vsize), uint(valign))
```
aleksisch: @borisbat What do you think, should we init it here? We do so in `AOT`, but maybe we can add a check related to existance of `finally` block.

PR #2071, opened 2026-02-16
jit: backtrace with function name

aleksisch:
Now stackwalk() in JIT is able to print function names. To do so `jit_emit_prologue` should be set in CodeOfPolicies.

But it won't print argument values. Since now it's pure binary code and there isn't much sense in converting all values back to das representation.


src/simulate/simulate.cpp
```diff
@@ -1559,6 +1559,7 @@ namespace das
             return showOutOfScope;
         }
         virtual void onCallAOT ( Prologue *, const char * fileName ) override {
+            // This also can be JIT.
```
aleksisch: @borisbat Can we determine based on Context whether we're in JIT or AOT? So we can print appropriate message here without adding extra flags etc.

PR #2431, opened 2026-04-17
compiler: allow return codes

aleksisch:
From now on daslang allows to return from main. And this value will be forwarded to daslang.exe process return code.

### DIFF
--- a/utils/daScript/main.cpp
+++ b/utils/daScript/main.cpp
@@ -339,13 +339,16 @@ int das_aot_main ( int argc, char * argv[] ) {
     return compiled ? 0 : -1;
 }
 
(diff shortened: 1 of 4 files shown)


utils/daScript/main.cpp
```diff
@@ -788,16 +796,17 @@ int MAIN_FUNC_NAME ( int argc, char * argv[] ) {
...
         return -1;
     }
 
     for ( auto & fn : files ) {
         replace(fn, "_dasroot_", getDasRoot());
-        if (!compile_and_run(fn, mainName, outputProgramCode, dryRun, compileOnly)) {
-            failedFiles++;
+        int rc = compile_and_run(fn, mainName, outputProgramCode, dryRun, compileOnly);
+        if ( rc != 0 ) {
+            exitCode = rc;
```
aleksisch, line 814: Note, that now we will return last error, instead of number of failed files.

PR #2434, opened 2026-04-17
fio: throw on invalid fopen mode instead of returning null

imp5imp5:
## Summary
- `builtin_fopen` now calls `context->throw_error_at(at, ...)` on invalid mode instead of silently returning `nullptr`. Silent failure was indistinguishable from "file not found" at the call site — any caller that forwards a bad mode will now get a daslang runtime error pointing at the offending call.
- Also throws on `NULL` name, matching the style of `builtin_fclose` / `builtin_fflush` / `builtin_fprint`.
- `is_valid_fopen_mode` collapsed to a one-liner using `strchr` + `strspn` (per review feedback).

## Test plan
- [x] New `tests/fio/fio_errors.das::test_fopen_invalid_mode` covers:
  - empty mode `""` throws
  - unknown first char `"q"` throws
  - unknown modifier `"rX"` throws
  - garbage mode `"zzz"` throws
  - valid mode `"rb"` on a missing file still returns null without throwing
- [x] `bin/Release/daslang.exe dastest/dastest.das -- --test tests/fio/` — 80/80 pass

### DIFF
--- a/src/builtin/module_builtin_fio.cpp
+++ b/src/builtin/module_builtin_fio.cpp
@@ -252,8 +252,18 @@ namespace das {
         if ( text ) fputs(text,(FILE *)f);
     }
 
+    static bool is_valid_fopen_mode(const char *mode) {
+        if (!mode || (mode[0] != 'r' && mode[0] != 'w' && mode[0] != 'a'))
+            return false;
+        for (const char *p = mode + 1; *p; ++p) {
+            if (*p != '+' && *p != 'b' && *p != 't' && *p != 'x')
+                return false;
+        }
+        return true;
+    }
+
     const FILE * builtin_fopen  ( const char * name, const char * mode ) {
-        if ( name && mode ) {
+        if ( name && is_valid_fopen_mode(mode)) {
             FILE * f = fopen(name, mode);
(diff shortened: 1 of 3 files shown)


src/builtin/module_builtin_fio.cpp
```diff
@@ -252,8 +252,18 @@ namespace das {
         if ( text ) fputs(text,(FILE *)f);
     }
 
+    static bool is_valid_fopen_mode(const char *mode) {
+        if (!mode || (mode[0] != 'r' && mode[0] != 'w' && mode[0] != 'a'))
```
aleksisch: Rewrite as:
```
  static bool is_valid_fopen_mode(const char *mode) {
      return mode && strchr("rwa", mode[0]) && mode[1 + strspn(mode + 1, "+btx")] == '\0';
  }
```

Or if you don't like `strspn`:
```
  static bool is_valid_fopen_mode(const char *mode) {                                                                                                                                               
      if (!mode || !strchr("rwa", mode[0]))
          return false;                                                                                                                                                                             
      for (const char *p = mode + 1; *p; ++p) {                                                                                                                                                     
          if (!strchr("+btx", *p))
              return false;                                                                                                                                                                         
      }                                                     
      return true;
  }

```

PR #2442, opened 2026-04-18
Feature/completely das aot

aleksisch:


src/ast/ast_simulate.cpp
```diff
@@ -3924,11 +3924,6 @@ namespace das
             logs << "unique        " << context.getUniqueMemorySize() << "\n";
         }
 
-        // log CPP
-        if (options.getBoolOption("log_cpp")) {
```
aleksisch, line 3928: @borisbat Do we still need `log_cpp`? It does aot and prints its content during simulate.

borisbat: nop. it can go away no problem

PR #2471, opened 2026-04-23
dasib: add option and result

aleksisch:
Add types for functional-style monadic computations.

PR #2454, opened 2026-04-21
add daslib/clargs

quasilyte:
Adds a command-line arguments handling library.
Also changes dastest das files
to use this new library.

Fixes #2316

PR #2482, opened 2026-04-25
clargs: short flags, --help renderer, standalone-tool args, Option<string> finders

borisbat:
## Summary

Follow-up to #2454. Adds the missing pieces of `daslib/clargs`:

- **`@clarg_short = "X"`** — single-character short flags, all field types, with collision detection and single-char validation (compile-time errors from the macro). Mixed long+short occurrences of array flags preserve declaration order.
- **`print_help` / `format_help`** — render `Usage` / `Flags` blocks from `CommandInfo`. Bool flags omit `=PLACEHOLDER`; enums render values inline; `(required)` and `(repeated)` markers in the doc column. `format_help` returns a string (testable, redirectable); `print_help` wraps it. `parse_args` does **not** auto-recognise `--help` — declare an explicit `help : bool` field and check it after parsing.
- **`get_program_args()`** — companion to `get_cli_arguments()` for AOT'd standalone tools that own argv directly (no `--` separator). Both helpers ship a `(argv)` overload to make the splitting logic unit-testable without touching the live process state.
- **`find_*_raw_value : Option<string>`** — replaces the `tuple<string, bool>` return shape with the freshly-merged `Option` type (#2471).

Also fixes a pre-existing bug in `collect_command_info`: `is_array` was being driven by `isVectorType()` (the `float2`/`int4`-style check), so `array<string>` fields silently came out with `is_array=false`. Switched to the actual `Type.tArray` check + `firstType` lookup.

`dastest` is migrated off the residual hand-rolled `--run` scan: `parse_args` already populates `test_args.run` from both `--run X` and `--run=X` forms, so the `find_index("--run")` + `args[runIdx+1]` dance is gone (also fixes the out-of-bounds when `--run` is the last token), and the "command-line args" error path returns `1` instead of `fio::exit(1)`.

Tutorial 53 + RST grow three new sections (short flags, help rendering, standalone-tool args) and the user-side `print_help` in the introspection section is replaced by a call to the library helper. `das2rst`'s `option` / `result` `Operators` groups extended to also catch `!=`, removing the pre-existing `Uncategorized` sections in the generated docs.

## Example

```das
[CommandLineArgs]
struct Config {
    @clarg_short = "n"  @clarg_required  @clarg_doc = "Greeting target"
    name : string

    @clarg_short = "v"  @clarg_doc = "Verbose output"
    verbose : bool

    @clarg_short = "h"  @clarg_doc = "Show this help and exit"
    help : bool
}

[export]
def main() {
    var cfg = Config()
    let err = parse_args(cfg)
    if (cfg.help) {
        print_help(get_command_info(type<Config>), "myprog")
        return
    }
    if (err != "") { print("error: {err}\n"); return }
    // ...
}
```

```
Usage: myprog [flags]

Flags:
  -n, --name=STRING    Greeting target (required)
  -v, --verbose        Verbose output
  -h, --help           Show this help and exit
```

## Test plan

- [x] `tests/daslib/clargs_test.das` — 34/34 pass (15 original + 19 new: short-flag matching per type, mixed long+short array order, required-flag short detection, metadata, `get_cli_arguments(argv)` / `get_program_args(argv)` testable overloads, `print_help` smoke)
- [x] Full suite: `bin/Release/daslang.exe dastest/dastest.das -- --test tests` → 6665/6665, no GC/handle/smart_ptr leaks
- [x] AOT regression: `bin/Release/test_aot.exe -use-aot dastest/dastest.das -- --use-aot --test tests` → 6077/6077
- [x] Lint clean on all changed files (one pre-existing failure on `dastest_clargs.das` — the lint tool's compile path doesn't recognise `@clarg_doc` field annotations; exists on master, separate bug)
- [x] Tutorial runs end-to-end: `bin/Release/daslang.exe tutorials/language/53_clargs.das` exits 0 with all 11 sections producing expected output
- [x] Sphinx clean build: zero warnings, zero errors
- [x] `format_file` clean on every changed `.das` file

### DIFF
--- a/daslib/clargs.das
+++ b/daslib/clargs.das
@@ -7,11 +7,13 @@ require daslib/templates_boost
 require daslib/strings_boost
 require strings public
 
+require daslib/option public
 require daslib/enum_trait public
 
 struct public CommandArgumentInfo {
     field_name : string
     flag_name : string
+    short_flag_name : string  // e.g. "-v"; empty if no short flag declared
     doc_string : string
     value_type : Type
     is_array : bool
@@ -23,8 +25,9 @@ struct public CommandInfo {
     args : array<CommandArgumentInfo>
 }
 
-def public get_cli_arguments() : array<string> {
-    var all_args <- get_command_line_arguments()
+// Returns the slice of `all_args` after a "--" separator (or empty if no "--").
+// Use this for script-style invocation (`daslang script.das -- script_args...`).
+def public get_cli_arguments(all_args : array<string>) : array<string> {
     var arg_index = find_index(all_args, "--")
     var result : array<string>
     if (arg_index == -1) {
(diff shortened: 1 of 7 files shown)


daslib/clargs.das
```diff
@@ -106,70 +135,291 @@ def public find_array_flag_raw_values(args : array<string>; flag_name : string)
...
+        dst = unwrap(opt)
+        return ""
+    }
+    if (flag_present(args, arg_info)) {
+        return "missing value"
     }
     return ""
 }
 
 def public parse_argument(var dst : int&; args : array<string>; arg_info : CommandArgumentInfo) : string {
```
aleksisch, line 239: @borisbat  Should be replaced to return `Result<int, string>`. We should never use arg by reference from now on.

daslib/clargs.das
```diff
@@ -106,70 +135,291 @@ def public find_array_flag_raw_values(args : array<string>; flag_name : string)
...
+    // Reject both outright parse failures and trailing garbage ("42abc").
+    if (result != ConversionResult.ok || offset != length(raw_value)) {
+        return "invalid int value '{raw_value}'"
+    }
+    dst = parsed
     return ""
 }
 
 
 def public parse_argument(var dst : float&; args : array<string>; arg_info : CommandArgumentInfo) : string {
```
aleksisch, line 260: `Result<float, string>`

daslib/clargs.das
```diff
@@ -106,70 +135,291 @@ def public find_array_flag_raw_values(args : array<string>; flag_name : string)
...
+    var offset = 0
+    let parsed = float(raw_value, result, offset)
+    if (result != ConversionResult.ok || offset != length(raw_value)) {
+        return "invalid float value '{raw_value}'"
     }
+    dst = parsed
     return ""
 }
 
 def public parse_argument(var dst : array<string>&; args : array<string>; arg_info : CommandArgumentInfo) : string {
```
aleksisch, line 279: `Result<array<string>, string>`

daslib/clargs.das
```diff
@@ -106,70 +135,291 @@ def public find_array_flag_raw_values(args : array<string>; flag_name : string)
...
         dst <- raw_values
+        return ""
+    }
+    if (flag_present(args, arg_info)) {
+        return "missing value"
     }
     return ""
 }
 
 def public parse_argument(var dst : bool&; args : array<string>; arg_info : CommandArgumentInfo) : string {
```
aleksisch, line 291: `Result<bool, string>`

PR #2484, opened 2026-04-25
option/result: support non-copyable payloads

borisbat:
## Summary

`Option<T>` and `Result<T, E>` previously broke at compile time on any non-copyable payload (`array<T>`, `table<K;V>`, lambdas) because every constructor / combinator / unwrap path used plain copy-assignment (`o._value = v`) and copy-return (`return o._value`).

This PR migrates both modules to dispatch on `static_if (typeinfo can_copy(...))`: copyable payloads keep the fast `=` / direct-return path; non-copyable payloads clone via `:=` (or move via `<-`) and `return <- clone_to_move(...)`. Behavior is identical for `int` / `float` / `string` / structs of those, and `array` / `table` / lambda-typed payloads now Just Work.

Adds three move-variants for genuinely move-only call sites:
- `move_some(var v)` — drains `v` into the option (mirrors `some` but moves)
- `move_ok(var v, type<E>)` and `move_err(var e, type<T>)` — same on each side of `Result`

The `==` / `!=` operators remain unchanged — they delegate to the inner type's `==`, so any payload with an `==` defined still works.

## What changed

- `daslib/option.das` — every read/return path migrated; added `move_some`. Plain `some(v)` now uses constructor-init `:= v` instead of post-init copy.
- `daslib/result.das` — same migration; added `move_ok` and `move_err`.
- `tests/option/test_option_non_copyable.das` (new) — 19 cases over `Option<array<int>>` covering every public function plus `move_some`.
- `tests/option/test_result_non_copyable.das` (new) — 22 cases over `Result<array<int>, string>` plus `move_ok` / `move_err`.
- `tests/README.md` — index entries.
- `doc/source/reference/tutorials/52_option_and_result.rst` — new "Non-copyable payloads" section, API tables include `move_*`.
- `tutorials/language/52_option_and_result.das` — runs an example showing clone vs move.
- `doc/source/stdlib/handmade/module-{option,result}.rst` — module preambles updated.

## Test plan

- [x] Interpreted: `bin/Release/daslang.exe dastest/dastest.das -- --test tests/option/` — 186/186 pass
- [x] AOT: `cmake --build build --config Release --target test_aot` succeeds; `bin/Release/test_aot.exe -use-aot dastest/dastest.das -- --use-aot --test tests/option/` — 186/186 pass
- [x] Lint: `mcp__daslang__lint` clean on all modified `.das` files
- [x] Format: `mcp__daslang__format_file` reports no changes needed
- [x] Tutorial 52 runs end-to-end with the new non-copyable example
- [ ] CI green

## Notes

- The auto-generated `doc/source/stdlib/generated/{option,result}.rst` aren't refreshed in this PR — that requires `DAS_HV_DISABLED=OFF` + `DAS_PUGIXML_DISABLED=OFF` and a `das2rst.das` run. The `//!` comments are accurate, so any future docs regen will pick the changes up.

### DIFF
--- a/daslib/option.das
+++ b/daslib/option.das
@@ -38,7 +38,14 @@ struct template Option {
 //! mutable source, use ``move_some`` instead.
 def some(v : auto(TT)) {
     typedef OT = $Option < TT - const >
-    return <- OT(_has_value = true, _value := v)
+    var o = default<OT>
+    o._has_value = true
+    static_if (typeinfo can_copy(o._value)) {
+        o._value = v
+    } else {
+        o._value <- clone_to_move(v)
+    }
+    return <- o
 }
 
 
--- a/daslib/option.das
+++ b/daslib/option.das
@@ -12,7 +12,13 @@ require daslib/typemacro_boost
 
 //! Monadic ``Option<T>`` type — represents a value that may or may not be present.
 //!
-//! Functional API for modelling optional value.
+//! Functional API for modelling optional value. The payload may be any type,
+//! including non-copyable ones (``array<T>``, ``table<K;V>``, lambdas):
+//! constructors and combinators clone or move on the non-copyable branch via
+//! ``static_if (typeinfo can_copy(...))``.
+//!
+//! Construct via ``some(v)`` (clones the payload), ``move_some(v)`` (moves out
+//! of a mutable source), or ``none(type<T>)``.
 //!
 //! See also ``daslib/result`` for a ``Result<T, E>`` counterpart.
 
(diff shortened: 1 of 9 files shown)


daslib/option.das
```diff
@@ -100,22 +140,38 @@ def unwrap(o : $Option < auto(TT) >) : TT {
...
+    }
 }
 
 //! Extract the payload or return ``d`` when ``none``.
 def unwrap_or(o : $Option < auto(TT) >; d : TT) : TT {
-    return o._has_value ? o._value : d
+    static_if (typeinfo can_copy(o._value)) {
+        return o._has_value ? o._value : d
+    } else {
+        return <- o._has_value ? clone_to_move(o._value) : clone_to_move(d)
```
aleksisch: ?
```suggestion
        return <- clone_to_move(o._has_value ? o._value : d)
```

daslib/option.das
```diff
@@ -100,22 +140,38 @@ def unwrap(o : $Option < auto(TT) >) : TT {
...
+    }
 }
 
 //! Extract the payload or return ``default<T>`` when ``none``.
 def unwrap_or_default(o : $Option < auto(TT) >) : TT {
-    return o._has_value ? o._value : default<TT>
+    static_if (typeinfo can_copy(o._value)) {
+        return o._has_value ? o._value : default<TT>
+    } else {
+        return <- o._has_value ? clone_to_move(o._value) : clone_to_move(default<TT>)
```
aleksisch:
```suggestion
        return <- clone_to_move(o._has_value ? o._value : default<TT>)
```

daslib/option.das
```diff
@@ -146,15 +206,27 @@ def zip(a : $Option < auto(TT) >; b : $Option < auto(UU) >) {
...
+    return <- r
 }
 
 //! Unwrap or default. Mirrors ``T? ?? default`` for value types.
 def operator ??(o : $Option < auto(TT) >; d : TT) : TT {
-    return o._has_value ? o._value : d
+    static_if (typeinfo can_copy(o._value)) {
+        return o._has_value ? o._value : d
+    } else {
+        return <- o._has_value ? clone_to_move(o._value) : clone_to_move(d)
```
aleksisch:
```suggestion
        return <- clone_to_move(o._has_value ? o._value : d)
```

PR #2578, opened 2026-05-05
Add tutorials to ci

aleksisch:


src/builtin/module_builtin_runtime.cpp
```diff
@@ -1532,7 +1532,7 @@ namespace das
     }
 
     char * builtin_shared_module_extension ( Context * context, LineInfoArg * at ) {
-#ifdef DAS_NO_ASSERTIONS
+#ifdef NDEBUG
```
borisbat, line 1535: its all over the code-base. very consistently DAS_NO_ASSERTIONS. so lets not make it inconsistent. either fix everywhere, or not at all

borisbat (changes requested):
personally i would not change everywhere.

src/builtin/module_builtin_runtime.cpp
aleksisch: That was the point of introducing DAS_NO_ASSERTIONS. To be able to enable assertions in Release mode. NDEBUG will be consistent with Debug build

borisbat: sure, lets fix the rest of the uses then? im thinking separate PR. file an issue for now, when stopping point - we go through and switch to NDEBUG

aleksisch: Changed in 2 more places.

Should we change this one as well? Everywhere else we use DAS_NO_ASSERTIONS for assertions.
```
#ifdef DAS_NO_ASSERTIONS
                error = context.getStackWalk(&debugInfo, false, false) + error;
#else
                error = context.getStackWalk(&debugInfo, true, true) + error;
#endif
```

PR #2624, opened 2026-05-10
daslib(delegate): public DelegateReturn template + dasImgui phase 0b.3 mouse cards

borisbat:
## Summary

- **`daslib/delegate.das`** — promote `struct template DelegateReturn` to `struct template public`. The void-return sibling `Delegate` was already public; the asymmetry was a copy-paste oversight that prevented cross-module instantiation of the non-void variant.
- **`mouse-data/docs/`** — 4 new mouse cards captured during dasImgui Phase 0b.3 (containers + hierarchical path keys), 1 obsolete card removed:
  - `dasimgui-container-annotation-and-hierarchical-path-keys` — how `[container]` differs from `[widget]` (same widget_ident injection / auto-emit gate, plus container_path_push/pop wrapping the body so leaf children get `MAIN/CHILD/LEAF` registry keys).
  - `daslang-block-last-arg-blocks-default-arg-fill` — `def foo(a, b=def, blk : block)` does NOT auto-shift defaults so BLOCK lands in slot 2; positional resolution is left-to-right and BLOCK collides with b. Workaround: drop defaults on per-call config for `[container]` defs, supply via the named-tuple at the call site.
  - `daslib-json-optional-bool-omitted-when-false-snapshot-test-impact` — `@optional bool : opened` is OMITTED from the JSON envelope when false; integration tests asserting on the false transition need to switch signal (e.g. leaf disappearance instead of `opened==false`).
  - `imgui-closable-window-skip-begin-when-closed-not-call-with-pfalse` — ImGui's contract: stop calling Begin once `*p_open` flips to false. Calling Begin with `*p_open=false` does NOT reliably suppress the body — leaves render at least one extra frame. dasImgui's `window` / `tab_item` containers gate the entire Begin/End on `state.open`.
  - `dasimgui-widget-auto-emit-skips-struct-field-defaults` — **deleted** (now-stale). Fixed upstream in dasImgui commit 111f7dc — the `[widget]` macro now emits `var IDENT : T = T()` so the default constructor runs and field defaults fire on the auto-emitted global.

## Test plan

- [x] daslib/delegate.das compiles + lints clean
- [x] Mouse index rebuild on the 4 added + 1 deleted cards completes (40 docs total)
- [ ] No code changes outside daslib/delegate.das — no test-suite impact expected


daslib/delegate.das
Y0L042: Is this not still `private`? I had issues with the void return delegate.

Y0L042: `struct template DelegateVoid` gives visibility errors. `struct template public DelegateVoid` works.

borisbat: Good catch — fixed. `DelegateVoid` is now `struct template public` (matching `DelegateReturn`).

daslib/delegate.das
```diff
@@ -52,7 +52,7 @@ def private build_invoke_call(at : LineInfo; var targetExpr : ExpressionPtr; fun
 
 // ── template: non-void return ──────────────────────────────────────────
 
-struct template DelegateReturn {
+struct template public DelegateReturn {
```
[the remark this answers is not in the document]
borisbat: Promoted `DelegateVoid` to `struct template public` to match `DelegateReturn` — same fix as Y0L042's thread.

PR #2645, opened 2026-05-13
dasOpenGL: async PBO ring for APNG capture

borisbat:
## Summary

Replace the sync `glReadPixels` in `record_tick` with a ring of N `GL_PIXEL_PACK_BUFFER` PBOs so the render loop never blocks on the GPU during APNG capture. Also drop the zlib compression level for the streaming encoder from stbi's default of 8 down to 4 — the encoder worker thread was the actual bottleneck at 1280x720+, and the level-4 tradeoff is right for tutorial-style recordings of bounded duration.

## Design

**Ring of PBOs** (default 4, range [2,8], per-recording override via `record_start` args):

- write side every tick: `glReadPixels` into `pbos[K%N]` with a null offset — returns immediately, GPU finishes asynchronously. Frame's `delay_ms` is stashed in a parallel ring slot.
- read side (once `frames_seen >= N`): `glMapBuffer` on `pbos[(K-N+1)%N]` (oldest in flight, GPU has long since finished), feed pixels to `stbi_apng_frame` with the matching delay, `glUnmapBuffer`.
- `record_stop` drains remaining N-1 in-flight buffers via the same `harvest_one_pbo` helper, then `glDeleteBuffers`.

Knob: `var public capture_pbo_count : int = 4` (module default, clamp [2,8]). `RecordStartArgs.pbo_count` per-recording override; `0` means use module default. `RecordStartResult` / `RecordStatusResult` report `pbo_count` so callers can see the effective ring size.

`screenshot` stays synchronous — single-shot, no benefit from amortization.

**PNG compression level 4** at `record_start`:

Level 8 caps the single encoder thread around 25-27 fps on a 1280x720 stream — the row-flip + filter + `stbi_zlib_compress` pipeline is CPU-bound on level 8. Level 4 is 2-3x faster per frame for a minor file-size cost. The comment at the call site flags this as a tutorial-friendly baseline so it's clear where to promote to a `compression_level` arg if a future caller wants lossless-ish capture.

## Numbers

Widgets-tour recording, 1024x720 / 30fps / 16.8s, on the master baseline vs this branch:

| Path | frames | dropped | drop rate |
| --- | --- | --- | --- |
| sync `glReadPixels` + level 8 | 443 | 128 | 29% |
| PBO ring + level 4 | 448 | 0 | 0% |

The PBO ring alone (level 8 retained) brought dropped count down modestly but the encoder still couldn't keep up — both knobs together get to zero drops.

## Test plan

- [x] Repeat-recording smoke at 1024x720 / 30fps — 0 dropped, 20 MB / 16.8s output
- [x] `record_stop` drains in flight buffers cleanly (saved frame count matches `frames_seen` modulo the final partial drain)
- [ ] Regression test (`test_record_pbo_drops.das`) — pending; encoder-bottleneck reframing made it less load-bearing. Will land alongside any future `compression_level` arg.

### DIFF
--- a/modules/dasOpenGL/opengl/opengl_live.das
+++ b/modules/dasOpenGL/opengl/opengl_live.das
@@ -168,6 +168,7 @@ struct RecorderState {
     pbos              : array<uint>
     pbo_delays_ms     : array<int>   // parallel ring: delay_ms captured at write time
     pbo_size          : int64
+    prev_png_level    : int      // stb_image_write's global PNG zlib level at record_start; restored at record_stop
 }
 
 var private recorder : RecorderState
@@ -212,37 +213,32 @@ def record_start(input : JsonValue?) : JsonValue? {
     // but we still validate the IDs before proceeding.
     var pbos : array<uint>
     pbos |> resize(n)
-    unsafe {
-        glGenBuffers(n, addr(pbos[0]))
-    }
+    glGenBuffers(n, unsafe(addr(pbos[0])))
     for (id in pbos) {
         if (id == 0u) {
+            glDeleteBuffers(n, unsafe(addr(pbos[0])))
+            delete pbos
             return JV((error = "glGenBuffers failed"))
         }
     }
     for (id in pbos) {
         glBindBuffer(GL_PIXEL_PACK_BUFFER, id)
-        unsafe {
-            glBufferData(GL_PIXEL_PACK_BUFFER, pbo_size, null, GL_STREAM_READ)
-        }
+        glBufferData(GL_PIXEL_PACK_BUFFER, pbo_size, null, GL_STREAM_READ)
     }
     glBindBuffer(GL_PIXEL_PACK_BUFFER, 0u)
 
     // Probe: stbi's default PNG compression level is 8 (high). That caps the
     // single worker thread to ~25-27 fps at 1280x720. Level 4 is the
     // tutorial-friendly sweet spot: 2-3x faster per frame at minor file-size
-    // cost. If keepers, promote to a record_start arg.
-    unsafe {
-        stbi_write_set_png_compression_level(4)
-    }
-    var writer_ptr : void?
-    unsafe {
-        writer_ptr = stbi_apng_begin(args.file, w, h, 4)
-    }
+    // cost. We capture the previous level and restore it in record_stop so
+    // unrelated PNG writes (screenshot, etc.) aren't affected. If keepers,
+    // promote to a record_start arg.
+    let prev_png_level = stbi_write_get_png_compression_level()
+    stbi_write_set_png_compression_level(4)
+    let writer_ptr = stbi_apng_begin(args.file, w, h, 4)
     if (writer_ptr == null) {
-        unsafe {
-            glDeleteBuffers(n, addr(pbos[0]))
-        }
+        stbi_write_set_png_compression_level(prev_png_level)
+        glDeleteBuffers(n, unsafe(addr(pbos[0])))
         delete pbos
         return JV((error = "could not open file", file = args.file))
     }
@@ -262,6 +258,7 @@ def record_start(input : JsonValue?) : JsonValue? {
     recorder.pbos             <- pbos
     recorder.pbo_size         = pbo_size
     recorder.pbo_delays_ms |> resize(n)
+    recorder.prev_png_level   = prev_png_level
     recorder_active           = true
     return JV(RecordStartResult(
         file        = args.file,
@@ -292,17 +289,13 @@ def private harvest_one_pbo() : bool {
         glBindBuffer(GL_PIXEL_PACK_BUFFER, 0u)
         return false
     }
-    var ok : int
-    unsafe {
-        ok = stbi_apng_frame(recorder.writer, ptr, recorder.width * 4, recorder.pbo_delays_ms[read_idx])
-    }
-    unsafe {
-        glUnmapBuffer(GL_PIXEL_PACK_BUFFER)
-    }
+    let ok = unsafe(stbi_apng_frame(recorder.writer, ptr, recorder.width * 4, recorder.pbo_delays_ms[read_idx]))
+    // glUnmapBuffer returns GL_FALSE if the driver invalidated the buffer
+    // contents while it was mapped — count it as a harvest failure so the
+    // frame doesn't slip in with junk pixels.
+    let unmap_ok = glUnmapBuffer(GL_PIXEL_PACK_BUFFER)
     glBindBuffer(GL_PIXEL_PACK_BUFFER, 0u)
-    if (ok == 0) {
-        return false
-    }
+    return false if (ok == 0 || !unmap_ok)
     recorder.frames_written++
     return true
 }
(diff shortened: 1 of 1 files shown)


modules/dasOpenGL/opengl/opengl_live.das
```diff
@@ -167,20 +186,67 @@ struct RecordStartResult {
...
+    }
+    let pbo_size = int64(w) * int64(h) * 4l
+
+    // Allocate the ring before opening the apng writer so a GL failure leaves
+    // no half-state behind. glGenBuffers can't really fail with a sane driver,
+    // but we still validate the IDs before proceeding.
+    var pbos : array<uint>
+    pbos |> resize(n)
+    unsafe {
+        glGenBuffers(n, addr(pbos[0]))
```
aleksisch: `unsafe(addr(pbos[0]))`. Let's add linter if unsafe block used for single expression? What do you think?

modules/dasOpenGL/opengl/opengl_live.das
```diff
@@ -167,20 +186,67 @@ struct RecordStartResult {
...
+    }
+    for (id in pbos) {
+        if (id == 0u) {
+            return JV((error = "glGenBuffers failed"))
+        }
+    }
+    for (id in pbos) {
+        glBindBuffer(GL_PIXEL_PACK_BUFFER, id)
+        unsafe {
+            glBufferData(GL_PIXEL_PACK_BUFFER, pbo_size, null, GL_STREAM_READ)
```
aleksisch: I don't think unsafe needed here at all. We need linter check for empty unsafe

modules/dasOpenGL/opengl/opengl_live.das
```diff
@@ -167,20 +186,67 @@ struct RecordStartResult {
...
+        }
+    }
+    glBindBuffer(GL_PIXEL_PACK_BUFFER, 0u)
+
+    // Probe: stbi's default PNG compression level is 8 (high). That caps the
+    // single worker thread to ~25-27 fps at 1280x720. Level 4 is the
+    // tutorial-friendly sweet spot: 2-3x faster per frame at minor file-size
+    // cost. If keepers, promote to a record_start arg.
+    unsafe {
+        stbi_write_set_png_compression_level(4)
```
aleksisch: same

modules/dasOpenGL/opengl/opengl_live.das
```diff
@@ -210,19 +281,63 @@ struct RecordStopResult {
...
+    if (ptr == null) {
+        glBindBuffer(GL_PIXEL_PACK_BUFFER, 0u)
+        return false
+    }
+    var ok : int
+    unsafe {
+        ok = stbi_apng_frame(recorder.writer, ptr, recorder.width * 4, recorder.pbo_delays_ms[read_idx])
+    }
+    unsafe {
+        glUnmapBuffer(GL_PIXEL_PACK_BUFFER)
```
aleksisch: same here, and everywhere around

modules/dasOpenGL/opengl/opengl_live.das
```diff
@@ -167,20 +186,67 @@ struct RecordStartResult {
...
+    // Allocate the ring before opening the apng writer so a GL failure leaves
+    // no half-state behind. glGenBuffers can't really fail with a sane driver,
+    // but we still validate the IDs before proceeding.
+    var pbos : array<uint>
+    pbos |> resize(n)
+    unsafe {
+        glGenBuffers(n, addr(pbos[0]))
+    }
+    for (id in pbos) {
+        if (id == 0u) {
```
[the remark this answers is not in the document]
borisbat: Fixed in 3023400. The early-return on `id == 0u` now calls `glDeleteBuffers(n, ...)` and `delete pbos` before returning the error, so a pathological driver doesn't leak GL buffers + the local array.

modules/dasOpenGL/opengl/opengl_live.das
```diff
@@ -167,20 +186,67 @@ struct RecordStartResult {
...
+    }
+    glBindBuffer(GL_PIXEL_PACK_BUFFER, 0u)
+
+    // Probe: stbi's default PNG compression level is 8 (high). That caps the
+    // single worker thread to ~25-27 fps at 1280x720. Level 4 is the
+    // tutorial-friendly sweet spot: 2-3x faster per frame at minor file-size
+    // cost. If keepers, promote to a record_start arg.
+    unsafe {
+        stbi_write_set_png_compression_level(4)
+    }
```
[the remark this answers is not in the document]
borisbat: Fixed in 3023400. `RecorderState` gains `prev_png_level`; `record_start` captures the level via `stbi_write_get_png_compression_level()` before setting to 4, and `record_stop` (plus the writer-open-failure cleanup) restores it. `screenshot` and any unrelated PNG writes are no longer permanently throttled.

modules/dasOpenGL/opengl/opengl_live.das
```diff
@@ -210,19 +281,63 @@ struct RecordStopResult {
...
+    var ok : int
+    unsafe {
+        ok = stbi_apng_frame(recorder.writer, ptr, recorder.width * 4, recorder.pbo_delays_ms[read_idx])
+    }
+    unsafe {
+        glUnmapBuffer(GL_PIXEL_PACK_BUFFER)
+    }
+    glBindBuffer(GL_PIXEL_PACK_BUFFER, 0u)
+    if (ok == 0) {
```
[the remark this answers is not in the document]
borisbat: Fixed in 3023400. `harvest_one_pbo` now captures `glUnmapBuffer`'s return; `GL_FALSE` means the driver invalidated the buffer contents while mapped, so it counts as a harvest failure rather than letting potentially junk pixels into the apng stream.

modules/dasOpenGL/opengl/opengl_live.das
```diff
@@ -210,19 +281,63 @@ struct RecordStopResult {
...
     return JV((error = "not recording")) if (!recorder_active)
+
+    // Drain any PBOs still in flight (frames seen but not yet read back).
+    // Best effort: a failed apng_frame here is logged via the dropped counter
+    // but doesn't block teardown.
+    while (recorder.frames_written < recorder.frames_seen) {
+        if (!harvest_one_pbo()) {
+            break
+        }
+    }
```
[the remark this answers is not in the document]
borisbat: Fixed in 3023400. The comment now correctly states that `stbi_apng_dropped()` only counts encoder queue overflows inside the writer (not GL-side harvest failures), so the resulting `frames` count is what actually made it to disk and a failed drain breaks the loop without surfacing in `dropped`.

modules/dasOpenGL/opengl/opengl_live.das
borisbat: Narrowed to `unsafe(addr(pbos[0]))` here in 3023400 — that's the minimum: `addr` of an array element is the only operation that needs the permission, the surrounding `glGenBuffers` call doesn't. A general lint for redundant `unsafe` block syntax (your broader suggestion) is a good followup but out of scope for this PR.

modules/dasOpenGL/opengl/opengl_live.das
borisbat: Confirmed — `glBufferData(target, size, null, usage)` doesn't need `unsafe` (the binding signature accepts the null-data variant without it). Removed in 3023400.

modules/dasOpenGL/opengl/opengl_live.das
borisbat: `stbi_write_set_png_compression_level` takes only `int level` — `unsafe` dropped here in 3023400.

modules/dasOpenGL/opengl/opengl_live.das
borisbat: Swept the whole file in 3023400. Dropped `unsafe` around: `glUnmapBuffer`, `stbi_apng_dropped`, `stbi_apng_end`, `glReadPixels(..., null)`, plus the two flagged above. Narrowed the remaining ones (`addr(pbos[0])` callsites and `stbi_apng_frame` with the mapped void*) from block to `unsafe(expr)` form. Only what genuinely needs the permission carries it now.

PR #2753, opened 2026-05-20
perf_lint: add PERF020 — redundant same-name workhorse cast

borisbat:
## Summary

Adds **PERF020** to `daslib/perf_lint`. Flags `T(x)` where `x` is already of workhorse type `T` — a no-op `ExprCall` against the cast-named overload that the typer does not elide.

Fires for the 15 workhorse cast names registered by `ADD_NUMERIC_CASTS` / `ADD_BITFIELD_CASTS` + `das_lexical_cast_*`:

`int`, `int8`, `int16`, `int64`, `uint`, `uint8`, `uint16`, `uint64`, `float`, `double`, `string`, `bitfield`, `bitfield8`, `bitfield16`, `bitfield64`.

```das
// Bad
def widen(a : int64) : int64 {
    return int64(a)                 // PERF020
}

// Good
def widen(a : int64) : int64 {
    return a
}
```

## Implementation

- **Match:** `call.func.fromGeneric?.name ?? call.func.name` is in the workhorse-cast set AND `arg._type.baseType` equals the cast's target type. Const/ref/temp qualifiers on the argument are ignored — only `baseType` matters.
- **Dispatch:** hoisted above the closure guard in `preVisitExprCall` so the rule fires anywhere (a redundant cast inside a closure body is still redundant). `perf_warning` handles macro-generated / generic-instantiation suppression.
- **Out of scope by design:** user-named bitfield/enum ctors (`MyBitfield(x)`, `MyEnum(x)`), vector ctors (`int2(x,y)` etc.), and `string(das_string)` (covered by PERF007 / PERF012). The single-argument gate excludes multi-arg ctors; user-named ctors don't match the workhorse name set.

## Drive-by

- `daslib/perf_lint.das:394` — dropped `uint64(intptr(...))`; `intptr` already returns `uint64`. Self-hit found by the new rule.
- `utils/lint/tests/perf019_int_cast_collapse.das` — `good_nested_cast` used `int(int(m1))` deliberately to exercise PERF019's nested-call unwrap, but that shape is now a textbook PERF020 hit. Replaced inner cast with an `int64` widen so the same PERF019 code path is exercised without tripping PERF020.

## CI

The lint-on-changed-files gate at [`extended_checks.yml:225-238`](https://github.com/GaijinEntertainment/daScript/blob/master/.github/workflows/extended_checks.yml#L225-L238) already runs the unified lint runner on every changed `.das` file in PRs. **No workflow edits needed** — the new rule rides the existing pipeline.

A daslib sweep found **~30 PERF020 hits in 14 other files** (aot_cpp.das, ast_boost.das, coverage.das, debug.das, profiler.das, others). Per the standard rule-then-sweep pattern, those go to a follow-up cleanup PR — they don't fail current PRs unless those files are touched.

## Test plan

- [x] `dastest` on `utils/lint/tests/perf020_redundant_cast.das` — 11 PERF020 fires, matches `expect 31208:11`
- [x] `dastest` on all `utils/lint/tests/` — 28/28 pass after PERF019 fixture restructure
- [x] Unified lint runner on `daslib/perf_lint.das` (paranoid+perf+style) — PASS
- [x] `das-fmt --verify` on the two changed `.das` files — Verified
- [x] Sphinx clean build — no new warnings on `lint.rst`
- [x] CI: `extended_checks` job will lint changed files; daslib sweep showed `daslib/perf_lint.das` (the only daslib file in this PR) is clean


daslib/perf_lint.das
```diff
@@ -917,6 +944,21 @@ class PerfLintVisitor : AstVisitor {
         self->perf_warning("PERF019: int(...) | int(...) on the same {kind} type can collapse to int(L | R) (one cast instead of two)", expr.at)
     }
 
+    // --- PERF020: T(x) where x is already T — redundant workhorse cast ---
+
+    def check_perf020_redundant_cast(call : ExprCall?) : void {
+        if (call == null || call.func == null || length(call.arguments) != 1) return
+        let fname = call.func.fromGeneric != null ? string(call.func.fromGeneric.name) : string(call.func.name)
```
aleksisch, line 951: `string(a ? b : c)` instead of `a ? string(b) : string(c)`? Can be added to linter too btw. It's not first time such code was written.

PR #2985, opened 2026-06-03
dastest: capture crashing subprocess stderr in isolated mode

borisbat:
## Problem

In `--isolated-mode`, dastest runs each test file in a worker subprocess and captures its output via `popen`, which reads **stdout only**. A hard crash (access violation / abort / `__fastfail`) prints its diagnostic to **stderr** — daslang's crash-handler C++ stack trace **plus** the daslang call stack (`src/hal/crash_handler.cpp`). That stderr was silently dropped, so a crashed isolated test reported only:

```
FAIL ... crashed (exit status 3)
```

— throwing away the single most useful artifact for diagnosing the crash.

## Fix

`run_iso_subprocess` now redirects the worker's stderr to a per-invocation temp file (`create_temp_file_result`), and **on abnormal exit** folds it into the crashed file's `log` so it's surfaced inline under the failing test.

Deliberately scoped:
- **Only on non-zero exit** — a clean run's stderr (e.g. libhv logs) is not attributed (no noise on green).
- **Separate file, not `2>&1`** — stdout (and the `##dastest##` JSON report protocol) is untouched, so stderr interleaving can't corrupt result parsing.
- Temp file removed after read.

## Before / after

Running an isolated test that does a wild-pointer write:

**Before:** `crashed (exit status 3)` — nothing else.

**After:**
```
=== RUN 'test_artificial_crash'
----- crashed subprocess stderr (crash handler / stack) -----
CRASH: EXCEPTION_ACCESS_VIOLATION (0xC0000005) at address ... writing address 0xdeadbeef
Stack trace:
  [ 0] das::SimNode_BlockNF::eval ...      ← full C++ stack
  ...
daslang stack ... test_artificial_crash from .../suite.das:468   ← + daslang call stack
FAIL ... crashed (exit status 3)
```

## Testing

- Verified end-to-end with an artificial wild-pointer-write test through `--isolated-mode`: the crash handler's full stack is now attributed to the failing test (and dastest's own stderr is empty); a passing isolated run is unchanged.
- Lint clean (`utils/lint/main.das`), pre-push format+lint hooks pass.

Motivated by debugging a CI-only Windows crash where the isolated worker died with no surfaced stack — this is the general fix so any handler-caught crash (AV/abort) shows its stack in the test output.

### DIFF
--- a/dastest/dastest.das
+++ b/dastest/dastest.das
@@ -180,8 +180,25 @@ def private build_iso_cmd(filesList : array<string>; prefix, suffix : string) : 
 // Run one subprocess and parse its streamed per-file JSON reports.
 def private run_iso_subprocess(cmd : string; timeoutSec : float; var br : BatchRun) {
     var pending : array<string>
+    // popen_timeout's stderr handling is timeout-dependent: with timeout<=0 it
+    // falls back to plain _popen (stdout only — stderr DROPPED), which is the
+    // path isolated mode hits since per-test test_timeout defaults to 0; with
+    // timeout>0 it routes stderr into the captured pipe, interleaved with the
+    // ##dastest## JSON markers. Either way a hard crash (AV / abort / __fastfail)
+    // writes its diagnostic — the crash-handler C++ stack + the daslang call
+    // stack — to stderr. Redirect stderr to a temp file (overrides both paths,
+    // and avoids the timeout>0 interleave) and, on abnormal exit, fold it into
+    // the crashed file's log so the stack is surfaced under the failing test.
+    let tfRes = create_temp_file_result("dastest_stderr_", ".txt")
+    let errPath = (tfRes is value) ? (tfRes as value) : ""
+    if (empty(errPath)) {
+        // The point of this is to NOT silently lose crash diagnostics, so make
+        // the temp-file failure visible rather than quietly dropping stderr.
+        br.err |> push("dastest: could not create a temp file for subprocess stderr capture ({tfRes as error}); stderr (crash stacks, etc.) will not be surfaced")
+    }
+    let runCmd = empty(errPath) ? cmd : "{cmd} 2>\"{errPath}\""
     unsafe {
-        br.exitCode = popen_timeout(cmd, timeoutSec) $(f) {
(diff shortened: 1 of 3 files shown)


borisbat:
Thanks for the review — addressed in the latest push (`ff99d53`):

**2. Misleading "popen drops stderr" comment** — fixed. Rewrote it to state the actual timeout-dependent behavior: `popen_timeout` with `timeout<=0` falls back to plain `_popen` (stdout only — stderr dropped), which is the path isolated mode hits since per-test `test_timeout` defaults to `0`; with `timeout>0` it routes stderr into the captured pipe, interleaved with the `##dastest##` markers. The comment now also notes the `2>file` redirect overrides both paths and avoids that interleave.

**3. No regression test** — added. New `_isolated_fixture_crash/test_crash.das` (hard wild-pointer crash → trips the OS crash handler → `CRASH:` + stack to the worker's stderr), and a new case in `test_isolated_mode.das` asserting the spawned dastest's output contains the `crashed subprocess stderr` header **and** the `CRASH` line. Also added a negative assert to the existing passing-suite case: a clean run must **not** emit the header. Verified locally end-to-end.

**1. Error message prints `{cmd}` not `runCmd`** — keeping `{cmd}` intentionally. That string is the user-facing "run this to reproduce" command; `runCmd` only differs by the appended `2>"<tempfile>"` redirect, whose temp path is already deleted by the time anyone reads the message. Including it would be noise, not signal. The actual test command is the useful part.

borisbat:
Round 2 (`0476ba5`) — both new comments addressed:

- **temp-file creation failure silently disables capture** → now pushes a clear message into `br.err` when `create_temp_file_result` fails (incl. the error), so it's surfaced rather than quietly dropping stderr.
- **header mislabels non-crash abnormal exits** → header is now neutral and carries the code: `----- subprocess stderr (abnormal exit {N}) -----`, folded in on any non-zero exit (crash / panic / compile error / timeout). Regression test + comment updated to match.

Verified locally (lint clean; crash fixture now shows `----- subprocess stderr (abnormal exit 3) -----` + the handler stack). Resolving the threads.

PR #3200, opened 2026-06-18
ci: give the mingw lane a compiler cache (sccache)

borisbat:
## Summary

`build_windows_mingw` is the long pole of PR CI (~25 min) — the **only Windows lane with no compiler cache**, so it recompiles the whole C++ tree from scratch every run. (It's also the only Windows lane building `dasClangBind` + LLVM, and it runs interpreter + JIT + AOT sweeps — but the missing *compiler cache* is the fixable part.)

Unlike clang-cl — which *must* use the VS generator and therefore can't use `CMAKE_<LANG>_COMPILER_LAUNCHER` (documented in its own job) — **mingw builds on Ninja**, so it can use sccache exactly like the `build` matrix:
- `sccache-action` + local-disk backend (`SCCACHE_GHA_ENABLED=false`);
- one `actions/cache` slot (`sccache-mingw-clang64-release`), restored read-only on PRs, refreshed on master after Build;
- launcher passed as `-D` on daslang's own configure (scoped, not leaked into the GLFW/libhv sub-builds).

## Rollout

The first **master** run after merge populates the slot (still a cold build); subsequent runs — **including PRs** — hit the cache. So this PR's own mingw run is still cold (the slot doesn't exist yet); it validates the wiring (`sccache --show-stats` at the end of Build), and the speedup lands once master seeds the cache.

## Files

- `.github/workflows/build.yml` — sccache wiring in the `build_windows_mingw` job only; no other lane touched.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -645,8 +645,28 @@ jobs:
           mingw-w64-clang-x86_64-ninja
           mingw-w64-clang-x86_64-openssl
           mingw-w64-clang-x86_64-glfw
+          mingw-w64-clang-x86_64-sccache
           bison
           git
+
+    # sccache (compiler cache) for the mingw lane. mingw builds on Ninja, which
+    # honors CMAKE_<LANG>_COMPILER_LAUNCHER (the VS-generator clang-cl lane can't,
+    # which is why only this lane gets it). Use the msys2-NATIVE sccache (installed
+    # above), not the hostedtoolcache one: a native-Windows sccache fails to
+    # CreateProcess msys2's clang.exe (the compiler's runtime DLLs aren't on
+    # sccache's PATH). Local-disk backend in one actions/cache slot, same approach
+    # as the build matrix. Fixed key (single config); PRs restore read-only, master
+    # refreshes the slot after Build.
+    - name: "Configure sccache (local-disk backend)"
+      shell: bash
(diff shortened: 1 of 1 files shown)


borisbat:
CI caught a real issue (not a flake), converting to draft pending a decision.

The sccache wiring **is** reached — Ninja invokes `sccache C:\a\_temp\msys64\clang64\bin\clang.exe …` — but it fails immediately on every TU with:

```
CreateProcess failed: The system cannot find the file specified.
```

This is the classic **native-sccache ↔ msys2-clang** mismatch: the sccache installed by `mozilla-actions/sccache-action` is a native-Windows binary, and when it tries to spawn msys2's `clang.exe` the compiler's msys2 runtime DLLs (libc++, libunwind, …) aren't resolvable in sccache's process environment, so `CreateProcess` reports "file not found".

**Likely fix:** install the **msys2-native** sccache via `setup-msys2`'s `install:` list (e.g. `mingw-w64-clang-x86_64-sccache`) and drop the `sccache-action`, so sccache runs inside the same msys2 environment as clang and resolves its DLLs. That needs a CI iteration to confirm the package name + that the integration actually caches — left for a deliberate pass rather than a speculative push. Everything else in the matrix's approach (local-disk backend, fixed cache slot, launcher scoped via `-D`) stays the same.

PR #3310, opened 2026-06-30
Prefill 4.2×: thread attention/requant/norm + size jobque to Apple Silicon performance cores

borisbat:
## Prefill 4.2× — thread the serial work + size the jobque to Apple Silicon performance cores

Once the Q8 GEMM reached llama.cpp parity (PR #3297/#3304), the dasLLAMA prefill ceiling was the **serial, single-threaded** non-GEMM work — attention (the only O(npos²) term), the activation requant, and the per-position rmsnorm. This branch threads all of it and fixes the worker-pool sizing that was silently starving it.

### Results — Llama-3.2-1B Q8, JIT, M-series (8 P-cores + 2 E-cores)

Prefill throughput (tok/s), vs the same model on the NEON-only llama.cpp CPU build at 8 threads:

| context | before | **after** | total | llama.cpp | ours/llama |
|--------:|-------:|----------:|:-----:|----------:|:----------:|
| 128 | 235 | **503** | 2.1× | 827 | 61% |
| 512 | 170 | **481** | 2.8× | 821 | 59% |
| 1024 | 121 | **416** | 3.4× | 765 | 54% |
| 2048 | 76 | **322** | **4.2×** | 670 | 48% |

The gap to llama.cpp closes from ~6× to ~2×. What remains widens with context — that's attention: llama.cpp uses flash-attention, ours is threaded-but-classic, so it pulls ahead as the sequence grows (a separate, future effort).

### What changed

1. **Thread the prefill attention over heads.** Every head runs the identical causal triangle (perfect load balance) and writes a disjoint output slice, so there's no shared scratch — each head gets its own score row. `softmax`/`softcap` gain fork-callable pointer cores (fork workers can only touch raw addresses, not main-context array objects); the array overloads forward to them, bit-identical.

2. **Vectorize the flat per-element loops.** The residual adds and the SwiGLU/GeGLU gate multiply were plain bounds-checked array loops the JIT does not auto-vectorize (an isolated bench shows the pointer + `[vectorize]` form is 5.3× faster). softmax now scales by the reciprocal (matching ggml's `ggml_vec_scale_f32`) and carries vectorize hints.

3. **Thread the activation requant and the rmsnorm** over positions (both independent per position) — `quantize_q8_0_into` and `rmsnorm` get fork-callable pointer cores. requant 9.1×, norm 11× on their buckets.

4. **Size the jobque worker pool to the performance cores on Apple Silicon** (`src/misc/job_que.cpp`). The pool used logical-cores-1 everywhere; on a P+E chip that spills a worker onto a slow efficiency core that becomes the straggler every `parallel_for` waits on, and the wasm-guard `DAS_MAX_HW_JOBS` default (4) caps a native build far below the real core count besides. It now queries `sysctl hw.perflevel0.logicalcpu` (gated on `hw.nperflevels >= 2`) and runs one worker per P-core — no -1 (the E-cores absorb the main thread + OS) and no cap. **Intel Macs and every other platform keep the existing cores-1 rule unchanged**; `DAS_JOBQUE_THREADS` still overrides. This is a general daslang runtime improvement for all `parallel_for` users on Apple Silicon, not dasLLAMA-specific.

### Correctness

Every step is verified bit-exact against running `forward()` once per position, and token-for-token against llama.cpp greedy decode (Llama-3.2-1B 40/40). `test_prefill` gains a 200-token case to cover the threaded attention arm. Full suites: 59/59 dasLLAMA, 175/175 jobque. Full local preflight green (17/17 gates: interp/JIT/AOT sweeps, docs, sequence).

### DIFF
--- a/modules/dasLLAMA/dasllama/dasllama_math.das
+++ b/modules/dasLLAMA/dasllama/dasllama_math.das
@@ -48,6 +48,27 @@ def vaxpy(d : float const?; s : float const?; scale : float; n : int64) {
     }
 }
 
+//! d[j] += s[j], width-8 vectorized (the residual-stream update). The plain `for k; a[k]+=b[k]`
+//! over bounds-checked arrays does NOT vectorize under the JIT — this pointer form is ~5x faster.
+[hint(unsafe_range_check, noalias = d, noalias = s)]
+def add_inplace(var d : float?; s : float const?; n : int64) {
+    for [vectorize, vectorize_width = 8, unroll_count = 2] (j in range64(n)) {
+        unsafe {
+            d[j] += s[j]
+        }
+    }
+}
+
+//! d[j] *= s[j], width-8 vectorized — the SwiGLU/GeGLU gate multiply (act(W1·x) * (W3·x)).
+[hint(unsafe_range_check, noalias = d, noalias = s)]
+def mul_inplace(var d : float?; s : float const?; n : int64) {
+    for [vectorize, vectorize_width = 8, unroll_count = 2] (j in range64(n)) {
+        unsafe {
+            d[j] *= s[j]
+        }
+    }
+}
+
 // n*d above this -> split the output rows across worker threads. Below it, the per-call
 // job-dispatch overhead (~90us) dwarfs the gain, so stay single-thread. (Measured on M1 Max:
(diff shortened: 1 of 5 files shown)


borisbat:
@copilot resolve the merge conflicts in this pull request

PR #3368, opened 2026-07-03
jobque team dispatch + dasLLAMA decode integration

borisbat:
## Summary

Adds **team-mode dispatch** to the job queue — a ggml-threadpool-shaped fork/join path where workers self-serve chunks off a single published atomic instead of one fifo job per chunk — and wires dasLLAMA's decode path onto it. Motivated by LLM decode, where a token is dozens of back-to-back small `parallel_for` matmuls and the per-chunk fifo push + wait-group + condvar wake dominates.

Headline: **1B decode 65.6 → 77.6 t/s** (~0.95× llama.cpp CPU at ctx 8 on M1 Max), forward emission +11%, decode math bit-exact (token-for-token unchanged).

## jobque team mode (C++ + daslib)

- `JobQue::setTeamMode(bool)` / `teamParallelFor(numChunks, work)` / `runTeamChunks` — sticky, opt-in. A dispatch costs **one seq bump**; workers claim chunk indices via a single `fetch_add`, the caller participates as a lane, and join is a spin on a remaining-counter. No fifo push, no mutex, no wait group, no condvar on the hot path.
- **Hybrid poll/park** (ggml's `--poll` shape): workers poll the team slot during the `set_jobque_worker_spin` window, then park; a publish that finds parked workers takes the fifo mutex and notifies. Zero mutex traffic while everyone spins. The sleep/wake race is closed by a seq_cst Dekker pair (`mParkedWorkers` ↔ `mTeamSeq`).
- Coexists with the fifo path — `parallel_for` keeps working while team mode is on (workers poll both).
- New builtins: `set_jobque_team_mode` / `get_jobque_team_mode` / `team_parallel_for_invoke`.
- New daslib: `team_parallel_for` + `TeamParallelForMacro` in `jobque_boost.das` — same lambda-clone infrastructure as `new_job`/`parallel_for` (clone once per worker, C++ side), rewrites to the `_invoke` builtin.

## dasLLAMA integration

- `maybe_parallel_for` routes through `team_parallel_for` when team mode is on (fifo `parallel_for` otherwise, inline below threshold).
- **Dispatch-lane alignment** (`get_dispatch_lanes()` = workers + caller): chunk counts were sized to the 7 workers, but 8 lanes serve them, so the last wave ran half-empty (28 chunks / 8 lanes = 87.5% util). Sizing to lanes took decode GEMV 109 → 118-120 GB/s (llama.cpp's exact repack-path rate). Applied at all dispatch sites; knob `set_dispatch_caller_lane` / `decode_prof --legacy-lanes`.
- **Attention threaded at any depth** under team mode — the fifo-measured crossover no longer applies at ~µs dispatch (1B ctx8 attn 420 → 194 µs/tok).
- **Fused q8 gate+up dispatch**: gate and up share one activation quantize + one region-list `groupn` dispatch into `ffn_gu`'s halves (bit-identical — same quants, same per-row dots, only the scheduling changes).
- Decode RoPE picks up the precomputed cos/sin table path (leaves called directly, dodging the pass-through-const-write-elision).
- Dispatch observability counters (`count_parallel_dispatch` / `count_inline_run`) surfaced in the forward profiler.

## Benchmarks / evidence

`bench_gemv_decode.das` + `matmul_variants.das` (12-variant GEMV race: unroll/ILP/fma-flush/dual-group/inline-scale) proved the decode "kernel top-end gap" was chunk-count misalignment, **not** the kernel — llama.cpp's live M1 GEMV (`ggml_gemv_q8_0_4x4_q8_0`) is the same shape as our `dot_q8q8_laneq4`. `emission_bench.das` is the short-prompt generation baseline (ttft + emit t/s vs llama.cpp). The perf-ledger entry in `API_REWORK.md` records the finding.

## Testing

- New `tests/jobque/test_jobque_team.das`: inline fallback, exactly-once coverage across chunk shapes, back-to-back dispatch storms (seq reuse + stragglers), fifo-while-team, mode toggling, fork-pool, production-shape.
- jobque suite **194/194 interp + 194/194 AOT**, JIT verifier-clean.
- Full `tests/` suite **10812/10818** (0 fail, 6 pre-existing skips) — the shared `job()` loop change is regression-free.
- Token-for-token parity **40/40 Llama-3.2-1B-Q8** (dense, fused gate+up) and **40/40 Gemma-3-1B-Q8** (RoPE-table, sliding-window, GeGLU) vs the llama.cpp oracle.
- Docs: das2rst regen clean, sphinx latex + html `-W` build succeeded (zero warnings).

## Follow-up (documented in API_REWORK.md ledger, NOT this PR)

fp16-inline-scale: llama.cpp's `block_q8_0x4` packs 4 fp16 scales with the 128 quant bytes (one stream, 136 B/group) vs our second fp32-plane stream (144 B). At the now-aligned ~118-120 GB/s the remaining cls/ffn per-op gap (1.04-1.12×) is almost exactly this byte ratio. The win needs an f16→f32 convert intrinsic + a repack variant + offset plumbing — its own arc.


borisbat:
Post-open fix pushed (`071dd82d2`): a 9-model benchmark sweep after opening this PR caught a **straggler epoch race** in the team-dispatch claim loop — the exit test compared the atomic `fetch_add` index against a *plain* `mTeamNumChunks` field, and the join only waits on `remaining`, so a straggler between its overflow `fetch_add` and the compare could pair an old-epoch index with the *next* op's chunk count. Consequences: bogus chunk executed through a dangling `mTeamWork` (SIGSEGV), stolen decrement + double-executed index, `remaining` going negative → next join spins forever. Loudest where K == lanes (decode attention): gemma-3-4b hung or crashed within one `emission_bench` run.

Fix: `mTeamChunkNext` + `mTeamNumChunks` are replaced by one 64-bit `mTeamOp` word `{numChunks:32 | counter:32}` — publish is a single seq_cst store, and a claim's `fetch_add` returns index **and** chunk count as a unit, so overflow self-describes and an index can never pair with another op's count. A valid claim's pending decrement blocks that op's join, which is exactly what keeps `mTeamWork` alive while dereferenced.

Validation: regression test added first (`team_epoch_storm` — double-exec + SIGSEGV on the old build within one `-jit` run); 500k back-to-back shape-storm iterations clean (old build: SIGSEGV inside 16k); jobque suite 195/195 interp + 195/195 AOT; full suite 10813/10819 (0 failed); parity 40/40 on Llama-3.2-1B / gemma-3-1b / gemma-3-4b; gemma-3-4b emission now completes at 0.91× llama.cpp.

borisbat:
tsan lane fix pushed (`3c3ad415d`): the reported race is **pre-existing fork-pool machinery**, not the team dispatch — `set_jobque_fork_pool` (main thread) writes the plain `Context::keepForkContexts` while a draining job's post-notify cleanup reads it in `releaseForkContext` on a worker (`parallel_for`'s join returns at `notify_and_release`, but the job wrapper's cleanup runs after the notify). The new test's block-to-block config toggles exposed it.

Fix: `keepForkContexts` / `forkSkipInitScript` / `forkSkipHeapReset` become `atomic<bool>` with relaxed loads/stores — either value is coherent at a toggle boundary (the fork either pools or deletes); it just must not be a data race. The Context copy-ctor never copied these, so no other fallout; relaxed loads are the same single-byte load on arm64/x64.

Verified with a local TSAN build (`-DDAS_USE_SANITIZER=tsan`): pre-fix **14 of 25** `test_jobque_team` runs flag the exact CI race, post-fix **0 of 25** and the full `tests/jobque` dir is TSAN-clean (195/195). Normal build: jobque 195/195 interp + AOT, JIT smoke clean.

PR #3485, opened 2026-07-16
Fix v_ldu over-read of by-value float3 in math builtins

virt00l:
v_ldu/v_ldu_p3(&v.x) load 16 bytes from a 12 byte float3. These arguments are by-value locals, so the optimizer knows the object extent and is free to act on the over-read: clang 21 folds every one of these to an unconditional trap under -fsanitize=object-size, and the same UB class has been observed silently dropping stores in release codegen in downstream engine bindings.

Load exactly 12 bytes with vec_loadu3, which vec3::operator vec4f and WrapVec3Arg already use. No behavior change: every consumer here reads xyz only, and on non-clang vec_loadu3 still lowers to v_ldu_p3.

### DIFF
--- a/include/daScript/simulate/aot_builtin_matrix.h
+++ b/include/daScript/simulate/aot_builtin_matrix.h
@@ -77,7 +77,7 @@ namespace das {
     inline float3 float3x4_mul_vec3p(const float3x4 &a, float3 b) {
         mat44f va;
         v_mat44_make_from_43cu_unsafe(va, &a.m[0].x);
-        return v_mat44_mul_vec3p(va, v_ldu(&b.x));
+        return v_mat44_mul_vec3p(va, vec_loadu3(&b.x));
     }
 
     inline float4 float4x4_mul_vec4(const float4x4 &a, float4 b) {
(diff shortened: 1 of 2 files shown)


aleksisch (commented):
Vecmath do have `v_ldu_p3_safe` with some complex implementation:
```
VECTORCALL VECMATH_FINLINE vec3f v_ldu_p3_safe(const float *m) { return vcombine_f32(vld1_f32(m), vld1_lane_f32(m + 2, vdup_n_f32(0), 0)); } // 2 loads + INS
```
Why not using it here instead of simple `memcpy`?

GCC has the same flag `-fsanitize=object-size` and it will keep failing after your fix I suppose since `v_ldu_p3` implementation is:
```
    __forceinline vec4f vec_loadu3(const float *v)
    {
#ifdef __clang__
      vec4f vv = v_zero();
      memcpy(&vv, v, sizeof(float) * 3);
      return vv;
#else
      return v_ldu_p3(v);
#endif
    }

NO_ASAN_INLINE vec3f v_ldu_p3(const float *m)
{
#if defined(DAGOR_TSAN_ENABLED) || defined(DAGOR_ASAN_ENABLED)
  return v_ldu_p3_safe(m);
#else
  return v_ldu(m);
#endif
}

```

virt00l:
Quick answer - because it's used in other (similar) places. 
We might need to change implementation of that API though.
And `v_ldu_p3` probably need to be changed too (make it safe by default), but that whole different story not directly related to this commit.
> 
> Why not using it here instead of simple `memcpy`?
> 
> GCC has the same flag `-fsanitize=object-size` and it will keep failing after your fix I suppose since `v_ldu_p3` implementation is:
> 
> ```

PR #3521, opened 2026-07-20
dasLLAMA: auto GPU tier config + resident classifier offload (+11-15%) + MTP/tier crash fix

borisbat:
Session goal was to find out whether anything MTP-related could beat our best non-MTP GPU decode baseline on `Qwen3.6-35B-A3B`. The answer is **no**, with a measured reason — but getting there turned up one real crash, one solid GPU win, and a config simplification.

All numbers: `decode_real_bench`, `tg-real256`, 8 prompts, 16 lanes, RTX 3060 Ti, in-process A/B (route flip only, so box drift cancels).

## 1. `auto` GPU tier config

Reproducing 26 tg took seven env vars, two of which are counts tuned to one box (`MOE_LAYERS=2`, `MOE_STREAM=35`). `DASLLAMA_GPU=1` (or `GpuTierWant.auto_tier`) now requests the measured-best set — `dn`/`attn`/`dense`/`shexp` on, `qkv` and `heat` off, both measured negative/wash in the S1+S3 decode-arc commits — and resolves the counts to `-1` = "ask for every layer", letting the upload walk's existing budget stops place the split.

On this box auto finds a **better** split than the hand-tuned one, with headroom left:

| config | split | VRAM | tg-real256 |
|---|---|---|---|
| explicit `MOE_LAYERS=2 MOE_STREAM=35 …` | 2 resident / 35 streamed | 2979MB | 26.11 ± 0.12 |
| `DASLLAMA_GPU=1` | 4 resident / 36 streamed | 3977MB | **26.30 ± 0.15** |

Both upload loops already drop a partial triple and break when VRAM or the pinned-mirror budget runs out, so this degrades on a smaller card rather than failing. Every `DASLLAMA_GPU_*` knob still overrides individually and arming is unchanged — existing invocations behave exactly as before. Not yet validated on other GPUs; the per-knob overrides are the escape hatch.

## 2. Resident GPU classifier route (`--cls-ab`)

The classifier plane (`vocab × dim`) is a decode step's largest single weight read — ~27% on A3B, where the trunk touches only 3B of 35B params but the classifier still reads 100% of itself. Measured **in isolation** (`DASLLAMA_GPU_CLS=1` now arms the tier by itself, so no expert layers are offloaded — `resident 0L`, 421MB VRAM, classifier only):

```
gpucls-off 15.68 ± 0.21  ->  gpucls-on 17.43 ± 0.19   (+11.2%, MTP file)
gpucls-off 14.55 ± 0.66  ->  gpucls-on 16.77 ± 0.61   (+15.3%, non-MTP file)
```

It composes with the resident shared expert rather than competing: `shexp-off 16.69 ± 0.38 -> shexp-on 17.57 ± 0.06` (+5.3% on top), for +20.8% cumulative over CPU-only.

Worth noting this is a **synchronous** offload that pays, which bounds the decode-arc rule: sync offload fails when the work is too small to amortize the submit (qkv 0, heat negative), not because it's synchronous. 413MB against a ~0.3ms submit is firmly on the paying side. It's also the cheapest possible use of a GPU — one fixed plane, no streaming, fits any card with ~500MB free.

Getting there needed three gates that all omitted `cls`: the arm condition, the upload condition (a separate "nothing requested" check), and a missing `g_moe_gpu_cls_route` A/B gate that every other tier surface already had.

## 3. Crash fix: MTP draft block vs the tier

`moe_gpu_layer_on_gpu()` was `l >= from_layer` with **no upper bound**. The MTP draft block sits at layer index `n_layers` — above every uploaded range — so it satisfied the test, routed to the tier, and hit:

```
EXCEPTION: vulkan tier: dispatch against a region that is neither resident nor streamed
           (fmt 1 element offset 10868490240)
```

Reproducible with any NextN model plus any tier offload. Nobody hit it because the MTP arc ran tier-off and the server ships `mtp = false`. Bounded by `g_moe_gpu_total_layers`, with `total == 0` keeping the old unbounded answer so the synthetic-stack tier tests are unaffected.

## 4. Batched MTP verify classifier

The 2-row verify streamed the whole classifier plane **twice** (`forward_prefill`'s tail for row 1, `mtp_verify_row0` for row 0). `mtp_verify_logits` now does one `rms_batch` + one `mm_b_f` covering every row:

```
batchcls-off 11.46 ± 0.07  ->  batchcls-on 11.89 ± 0.01   (+3.8%)
```

Below the ~10–17% a fully eliminated stream is worth, so `mm_b_kq` recovers only part of it at `npos=2`. Output-invariant — accept counts identical across the flip, and `test_mtp` passes including the K-quant 27B case that covers `mm_b_kq`.

`forward_prefill`/`_body` take `skip_logits` (defaulted, all existing callers byte-identical); when set, the tail returns after the `rms` + `mtp_h` stash so the caller owns the classifier. `mtp_cls_batch_ok` keeps tied-fp32 and GPU-resident classifiers on the per-row path — neither has a batched form.

## Why the MTP-side change is deliberately small

Under the full tier, MTP measures **0.45x** (`mtp-off 27.91` vs `mtp-on 12.55`, 70.9% accept) — a spec step costs 3.80 decode-equivalents. The new `verify_batch_probe` isolates why:

```
decode (forward)  : 36902 us/position
prefill width 1   : 72178 us/step   (1.96x decode)   path overhead 96%
prefill width 2   : 90189 us/step   (2.44x decode)
prefill width 8   : 157236 us/step  (4.26x decode)
```

Identical work through `forward_prefill` instead of `forward` costs **96%**, because every GPU decode win (deltanet decode-step chain, resident shared expert, GPU classifier) lives on the `forward` path only, and a verify uses the prefill path. At depth 1 that's fatal in a way no drafter can fix: `verify(2) = 90189` exceeds `2 × decode = 73804`, so speculation loses even at 100% acceptance.

Speculation only regains room at width ≥ 4 (`3.10 < 4`, `4.26 < 8`), which needs recursive drafting plus routing narrow verifies through the decode chain. Left as future work — at 78% acceptance even width 8 pencils out near 0.67x.

## Benches

- `verify_batch_probe.das` (new) — measures `b`, the marginal cost of an extra verify row, the quantity that decides whether any speculative scheme can pay. Includes a width-1 control isolating path-switch cost from batching.
- `decode_real_bench.das` — `--mtp-ab`, `--batchcls-ab`, `--cls-ab`, `--grouped-off`.

## Gates

- `test_mtp` 8/8 (incl. K-quant 27B invariance)
- `test_vulkan_tier` 16/16
- lint + format clean on all changed files

### DIFF
--- a/modules/dasLLAMA/benchmarks/decode_real_bench.das
+++ b/modules/dasLLAMA/benchmarks/decode_real_bench.das
@@ -60,6 +60,18 @@ struct Args {
     @clarg_doc = "Run the timed reps twice in-process — attention q/k/v group OFF then ON (needs DASLLAMA_GPU_QKV=1 for the uploads)"
     qkv_ab : bool
 
+    @clarg_doc = "Run the timed reps twice in-process — MTP/NextN self-speculation OFF then ON (needs a NextN head in the GGUF; the ON arm also reports draft accept rate)"
+    mtp_ab : bool
+
+    @clarg_doc = "Force the per-position MoE prefill instead of the grouped buckets (bit-identical, pure perf lever) — the grouped path is tuned for large npos, so this probes whether it loses at the MTP verify's npos=2"
+    grouped_off : bool
+
+    @clarg_doc = "Run the timed reps twice in-process with MTP ON both times — batched verify classifier OFF then ON. Drift-cancelled measurement of the one-classifier-stream-per-verify win"
+    batchcls_ab : bool
+
+    @clarg_doc = "Run the timed reps twice in-process — resident GPU classifier plane OFF then ON (needs DASLLAMA_GPU_CLS=1, which arms the tier by itself so the classifier offload is measured alone)"
+    cls_ab : bool
+
     @clarg_short = "?"
     @clarg_doc = "Show this help and exit"
     help : bool
@@ -148,6 +160,9 @@ def main() {
     if (cfg.reuse) {
         set_expert_reuse(true)
     }
+    if (cfg.grouped_off) {
+        set_moe_grouped_prefill(false)
+    }
     var tk <- load_tokenizer_auto(cfg.model)
     print("loading {cfg.model}\n")
     with_job_que() {
@@ -162,7 +177,7 @@ def main() {
         }
         t.config.seq_len = min(t.config.seq_len, maxlen + ngen + 8l)
         let c = t.config
-        print("config: dim={c.dim} layers={c.n_layers} experts={c.n_expert} k={c.n_expert_used} | backend {active_kernel_backend()} | t{threads} affinity {pinned ? "hard" : "env/off"} | {length(toks)} prompts x {ngen} tok x {reps} reps\n")
+        print("config: dim={c.dim} layers={c.n_layers} experts={c.n_expert} k={c.n_expert_used} | backend {active_kernel_backend()} | t{threads} affinity {pinned ? "hard" : "env/off"} | moe-prefill {get_moe_grouped_prefill() ? "grouped" : "per-pos"} | {length(toks)} prompts x {ngen} tok x {reps} reps\n")
         // loud tier line — a silent GPU no-engage (absent DLL, failed probe) must never pass as a row
         let gs = gpu_tier_status()
         let gname = empty(gs.backend) ? "OFF" : "{gs.backend} {gs.device}"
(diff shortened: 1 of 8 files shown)


borisbat:
All three Copilot comments accepted — pushed in `63dbc61ec`.

**The two `-1` sentinel comments were a real bug, not a nit.** Both "nothing requested" gates test the counts with `<= 0`, so the new auto sentinel read as *off*. A want of `GpuTierWant(moe_layers = -1)` with every other rail down would have silently failed to arm — precisely the silent no-engage this code panics over elsewhere.

Worth recording why my own verification missed it: `DASLLAMA_GPU=1` also turns `dn` on, which armed the tier by a second route and masked the broken count gate entirely. The passing test proved the wrong thing. Both gates now test `== 0`, and I re-verified **both** directions rather than just the happy path:

```
DASLLAMA_GPU=1 -> vulkan RTX 3060 Ti | resident 4L streamed 36L dn 30 attn 10 dense 110 cls on 3977MB
no env         -> gpu OFF | resident 0L ... | not riding: no GPU tier installed
```

The panic-path nit is fixed in `decode_real_bench.das` as well — `verify_batch_probe`'s `load_corpus` was copied from it, so both were pathless.

Also folded in while here (not from review): the server and control page no longer hardcode `K=2 / S=35`. Those were tuned to one 3060 Ti, and auto measured *better* on that same box (4/36 → `tg-real256` 26.30 vs 26.11). `0` now means auto, which additionally removes the control-page trap where saving the raw unset sentinels wrote explicit zeros and disarmed the tier on restart. The server's blessed shape also gains `shexp = true` — the resident shared expert measured +8.3% in decode-arc S2 and the server was never setting it.

PR #3526, opened 2026-07-21
dasLLAMA: GPU tier serves mixed q8/kq expert triples (opens glm4moe)

borisbat:
Follow-on to #3521 — this commit was pushed to that branch after the merge snapshot was taken, so it didn't land. Self-contained: opens the GPU expert rail for `glm4moe` (and any other model llama.cpp's Q4_K recipes touch).

## The problem

llama.cpp's Q4_K MoE recipes bump `ffn_down_exps` to Q5_0/Q8_0 while `gate`/`up` stay Q4_K. After the loader's exact Q5_0→Q8 transcode, **every** layer's expert triple is `k4/k4/q8`. The upload walk demanded `all_q8 || all_kq`, so on GLM-4.5-Air it broke at the *first* layer and no expert stack could ever be tier-resident — with 2.8GB of a 4.2GB budget unspent.

This is not a VRAM problem and not model-specific: it's a property of the quantizer. Gemma hits the same shape, and **both** GLM quants mix (`Q4_K_M`: down = Q5_0×24 / Q8_0×22; `Q4_K_S`: Q5_0×42 / Q5_1×4), so "use a different quant" was not an escape — I downloaded Q4_K_S to check, and it doesn't even load (Q5_1 unsupported).

## The device side was already mixed-capable

Both `ffn_cmd_for` and `record_ffn_batch_cmd` already bind per-stack pipelines (with an explicit rebind when `gate != up`) and select the requant pipeline **off the down stack** — `actrq_k_pipeline` vs `actrq_pipeline`, with the matching `256:32` dispatch. Their comments state the intent outright. Only host-side guards blocked it.

## The fix

Split the single `xunit` into two. `gate`/`up` genuinely must agree — `fill_stack_acts` hands both the **same** `xqp` buffer, so one activation form. `down` consumes the *intermediate*, a separate activation, so it gets its own unit:

- `ffn_gemv_prep` (decode) and `vk_moe_ffn_batch` (batch), plus `ffn_chunk_rows`, whose chunk cap sizes the input window and the intermediate planes against different buffers
- the resident **and** streamed upload walks now require only that `gate`/`up` share a class

## Measured — GLM-4.5-Air Q4_K_M (RTX 3060 Ti, 16 lanes, `tg-real64`)

| config | VRAM | t/s | Δ |
|---|---|---|---|
| tier off | 0 | 5.20 | — |
| + classifier | 514MB | 5.45 | +4.8% |
| + shared expert | 1359MB | 5.72 | +5.0% |
| **+ 1 expert layer** | 2997MB | **5.80** | **+1.4%** |

+1.4% is small because one 1.85GB triple is all the 4.2GB cap holds. **The point is the slope**: each resident layer is `1/46` of routed-expert traffic, so this model now scales with VRAM instead of being capped at classifier+shexp forever. Extrapolating the measured slope (`t(L) ≈ 175ms − 2.24ms·L`), 10 t/s would need ~34 resident layers ≈ 64GB.

## Tests

New `k4/k4/q8` arm in `test_vulkan_tier` with **both** decode and batch (40-row) cells against the CPU reference; `ref_ffn_kq` takes a `midblk` so the intermediate requant matches the down form.

Red-then-green verified by reverting only the two engine files — the arm trips `q8/kq mixed stack triple in one FFN chain` at `dasllama_math_vulkan.das:3652` without the change:

```
without engine change : 18 tests, 16 passed, 2 errors
with engine change    : 18 tests, 18 passed, 0 errors
```

`test_mtp` 8/8 unaffected. Lint + format clean.

## Known gaps (recorded, not fixed here)

- **auto doesn't clamp `nstream`** against the tier's slot reserve — 46 layers throws `stream slots exceed STREAM_RESERVE`. Auto clamps *resident* layers via the walk's budget stops, but nothing bounds streaming against tier capacity. Worked around in measurement with `DASLLAMA_GPU_MOE_STREAM=0` (streaming is prefill-only anyway — the dense/stream rails gate on `npos >= 32`).
- **No unit coverage on the upload-walk gate.** `test_vulkan_tier` uploads synthetic stacks directly and never enters `moe_gpu_upload_resident`, so the `dasllama_common` half is exercised only by the real-model run.
- **Q5_1 unsupported**, which is what blocks Q4_K_S entirely.
- **No text-level correctness check** for GLM with the tier engaged — the unit arms validate the kernel against a CPU reference, which is not the same thing.

### DIFF
--- a/modules/dasLLAMA/dasllama/dasllama_common.das
+++ b/modules/dasLLAMA/dasllama/dasllama_common.das
@@ -2059,12 +2059,13 @@ def moe_gpu_upload_resident(var t : Model) {
         let f1 = fmt_at(t.we1_fmt, l)
         let f2 = fmt_at(t.we2_fmt, l)
         let f3 = fmt_at(t.we3_fmt, l)
-        let all_q8 = f1 == KqFmt.q8 && f2 == KqFmt.q8 && f3 == KqFmt.q8
-        let all_kq = moe_gpu_fmt_kq(f1) && moe_gpu_fmt_kq(f2) && moe_gpu_fmt_kq(f3)
-        // dense-lead layers carry no expert stacks; the tier offloads plain-q8 or all-K-quant
-        // stack triples (a q8/kq mix would need two activation forms in one chain) — either way
-        // the walk stops so the resident set stays contiguous from the end
-        if (l < t.config.n_layer_dense_lead || t.we1_offs[l] < 0l || !(all_q8 || all_kq)) {
+        // gate/up must share a form (one activation image); down is free. llama.cpp's Q4_K_M MoE
+        // recipe bumps ffn_down_exps to Q5_0/Q8_0, so k4/k4/q8 is the COMMON shape (glm4moe,
+        // gemma). The walk stops on refusal so the resident set stays contiguous from the end.
+        let gu_ok = ((f1 == KqFmt.q8) == (f3 == KqFmt.q8)
+            && (f1 == KqFmt.q8 || (moe_gpu_fmt_kq(f1) && moe_gpu_fmt_kq(f3))))
+        let dn_ok = f2 == KqFmt.q8 || moe_gpu_fmt_kq(f2)
+        if (l < t.config.n_layer_dense_lead || t.we1_offs[l] < 0l || !(gu_ok && dn_ok)) {
             break
         }
         var got = 0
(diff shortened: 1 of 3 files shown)


borisbat:
Both accepted — pushed in `22f419120`.

**The `midblk` one is the more valuable catch.** A reference that silently drops `h`'s tail wouldn't just be inaccurate, it would compute a *wrong* baseline and could mask a genuine kernel failure — the exact opposite of what the arm exists to do. Guarded now.

One deliberate change from the suggestion: I used **`verify`, not `assert`**. Per this project's conventions `assert` may be compiled out while `verify` stays in release — and a guard that vanishes in the build the tests actually run under would have answered the review without holding in practice.

The cleanup one is a straight fix: the arm now frees its stacks, dequantized weights and activations, matching the kq arm's convention.

`test_vulkan_tier` 18/18, lint and format clean.

PR #3541, opened 2026-07-22
dasSpirv coopmat: the native INT8 path (s8 x s8 -> s32)

borisbat:
## dasSpirv coopmat: the native INT8 path (s8 × s8 → s32)

Follow-up to the coopmat emitter (#3539). Adds the **int8** cooperative-matrix tile — the one that matters for quantized inference. **Q8_0 weights are int8**, so this feeds them straight into the Tensor Cores with **no dequant step** (the hand-written subgroup GEMM pays a dequant-to-f16 tax the tensor-core int8 path skips entirely).

### Surface

Marker structs for the **16×16×32** tile (K=32, double the f16 tile's MACs — A is 16×32, B is 32×16, s32 accumulator 16×16):

```
var a   : coopmatA_s8_16x32
var b   : coopmatB_s8_32x16
var acc : coopmatAcc_s32_16x16
coopmatLoad(a, ssbo_a_i8, off, stride, 0)
coopmatLoad(b, ssbo_b_i8, off, stride, 1)
acc = coopmatMulAdd(a, b, acc)          // emits the signed-components mask
coopmatStore(acc, ssbo_c_i32, off, stride, 1)
```

### Emitter

- `coopmat_info` now carries the component **kind** (float vs signed-int + width); `coopmat_type_of` builds an `OpTypeInt` component for the int tiles.
- The MulAdd arm appends a `CooperativeMatrixOperands` mask (`A|B|C|Result` signed = **0xF**) for an integer accumulate. The float path takes no mask and is **byte-identical** — every existing golden regenerates the same.
- The int8 SSBO tiles pull `StorageBuffer8BitAccess` via the existing 8/16-bit storage wave — no new machinery.

### Verified on hardware (RTX 3060 Ti, VK_KHR_cooperative_matrix)

The s8→s32 tile is **bit-exact** against an integer CPU reference, and runs at **~30 TOP/s** vs the f16 tile's ~19 TFLOP/s and the hand-written subgroup GEMM's ~4.2 (rough, uncleared box). The device advertises `s8×s8→s32` at 16×16×32 among its 10 configs, which is what this targets.

### Tests

`coopmat_i8` fixture (s8×s8→s32); `test_coopmat` pins the opcodes + the `StorageBuffer8BitAccess` capability + **the signed-components mask** (operand 5 == 0xF) + spirv-val (vulkan1.3); census **pass-through** (no new opcodes over the f16 fixture); a byte-identical golden. All existing coopmat/golden/census tests unchanged (0 existing goldens modified).


borisbat:
Re: the two Copilot notes on the untyped `coopmatLoad`/`coopmatStore` buffer params — the untyping is deliberate, not a lapse, and typed `@ssbo` overloads would reject valid usage:

The same coopmat marker family legitimately backs **three different element types**: `array<int8>` (the standalone GEMM probe), `array<uint>` packed 4×int8-per-word (the real dasLLAMA `q8_moe_batch` weight/act planes), and `T[N]` `@workgroup` staging tiles (the shared-memory load/store path this PR adds). All three feed the *same* `coopmatA_s8_*` / `coopmatB_s8_*` marker, so a typed `array<int8>` overload would type-*reject* the uint-packed and workgroup cases that are the actual integration targets.

The emitter consumes the buffer arg structurally via `ptr_of` (never `value_of`) and keys the load/store lowering on the coopmat marker's origin name, not the buffer element type — so element-type is not a meaningful compile-time constraint here. Validity is enforced downstream by the three-layer net the quartet established: `validate_spirv` (spirv-val) in the fixture, the `spirv-dis`-vs-glslang golden diff in CI, and the on-3060 Ti dasVulkan parity test. Merging; will revisit if a real mistyping slips past all three.

PR #3582, opened 2026-07-27
Fix Windows probes in the dasllama bench harness

artarasov:
llama-bench --ref never spawned on Windows: _popen goes through cmd /c, which strips the first and last quote of the command line, so the quoted binary path breaks. das_version probed bin/daslang, which cmd can't resolve, and wmic is gone on Win11 24H2+ so the ram/gpu fields came back empty. All three fixed in profile_common.das. Found while benching SmolLM2 on a 14700HX box.


borisbat (commented):
Verified all three fixes on a Windows 11 Pro 26200 (25H2) box, through the real `popen` path rather than from a shell:

| Claim | Result |
|---|---|
| `cmd` can't resolve `bin/daslang` | confirmed — `'bin' is not recognized`; `bin\daslang` → `0.6.3` |
| `cmd /c` strips the first and last quote | confirmed — `"…\where.exe" "cmd.exe"` → `'C:\Windows\System32\where.exe" "cmd.exe' is not recognized` |
| The sacrificial outer wrap fixes it | confirmed |
| The wrap is "safe either way" | confirmed on a plain command, on one with a `2>nul` redirect, and on a single bare token |

Lint and format are clean on the changed file. Diagnoses all look right. One thing I think needs another pass before this lands.

## The virtual-adapter preference sits in the branch that doesn't run

The premise is that wmic is gone on 24H2+. It isn't gone on this box — 25H2, and wmic answers fine:

```
Name=Parsec Virtual Display Adapter

Name=NVIDIA GeForce RTX 5060 Ti
```

`platform_gpu()` returns the **first** `Name=` line and short-circuits (`profile_common.das:483-488`, unchanged by this PR), so this box records `Parsec Virtual Display Adapter` as its GPU — and the new prefer-physical logic below it never executes. As written the ordering fix only helps boxes where wmic is actually absent, which is not the same set as "24H2+" (the WMIC feature-on-demand is deprecated and off by default on clean installs, but survives an in-place upgrade).

Suggest hoisting the selection into a small helper and running it over both outputs, or dropping the wmic branch entirely and always going through CIM.

## Smaller things

- **`"Virtual"` is a narrow filter.** It catches Parsec and VirtualBox, but not `Microsoft Remote Display Adapter`, `Microsoft Basic Display Adapter`, `Microsoft Hyper-V Video`, `VMware SVGA 3D`, or `Citrix Indirect Display Adapter`. Case-sensitive as well.

- **The wmic miss is noisy.** `run_capture` doesn't redirect stderr, so on a genuinely wmic-less box the harness prints `'wmic' is not recognized as an internal or external command,` to the console twice in the middle of a bench run. `2>nul` on both wmic commands handles it — and it survives the popen path (tested above).

- **The shell-free path already exists in-tree.** `daslib/fio.das:648 run_and_capture(args, output, timeout)` goes through `popen_argv` → `CreateProcess` directly, so neither the quote rule nor `/` vs `\` applies. `das_version()` converts in one line and the new `is_windows()` branch disappears with it. `run_capture` can't move wholesale — it runs real shell pipelines like `lspci | grep` — and `run_llama_bench` would need its three callers to build argv, so the sacrificial wrap is a reasonable pragmatic fix there. Worth naming `popen_argv` in that comment so the next person doesn't re-derive the `cmd /?` rule.

- **`run_capture` has no `f != null` guard** (`profile_common.das:307`) while `run_llama_bench` does. Pre-existing, but this PR is specifically about probes failing to spawn.

Thanks for chasing this down — the `cmd` quote rule in particular is a trap we keep re-hitting.

artarasov:
Applied the review in bab2c32f0:

- **The GPU pick now runs over both probe outputs.** The wmic branch collects every `Name=` line instead of returning the first, CIM fills the list only when wmic yields nothing, and one shared `pick_gpu_name` prefers the first non-virtual adapter with fallback to the first name — so a 25H2 box with wmic present and a Parsec adapter enumerated first now records the physical GPU.
- **Broader virtual filter.** Case-insensitive, matches `virtual`, `remote display`, `basic display`, `hyper-v`, `vmware`, `citrix`. All-virtual and single-Basic-Display lists still fall back to the first name rather than reporting nothing.
- **`2>nul` on both wmic probes**, and `run_capture` got the same `f != null` guard `run_llama_bench` already had.
- **`das_version` goes through `run_and_capture`** — with one correction to the "converts in one line" note: `popen_argv` is only half shell-free on Windows. `CreateProcessA` with a NULL application name resolves `bin\daslang` but not `bin/daslang` (probed both: forward slash spawns nothing and returns -1, backslash prints 0.6.4). So the conversion keeps a small Windows branch that normalizes `/` to `\` before spawning — which as a bonus fixes a forward-slashed `DASLANG_BIN` override, something the old cmd path also broke on. The caveat is now in `run_and_capture`'s docstring in `daslib/fio.das` so the next caller doesn't rediscover it.
- **The sacrificial-wrap comment names `popen_argv`** as the shell-free alternative and why it doesn't apply there.

Verified on a Windows box: lint and format clean on both files; `das_version()` returns 0.6.4 through the argv path with both the default and a forward-slashed override; `pick_gpu_name` unit-tested against Parsec-first, RDP-first, all-virtual, single-Basic-Display, and empty adapter lists; `gather_hardware` end-to-end reports the physical GPU and RAM through the `2>nul` wmic path.

PR #3597, opened 2026-07-31
dasSpirv: lower a daslang enum to its underlying integer type

olegus8:
Lowers a daslang enum to the `OpTypeInt` of its underlying integer, so a shader
can use enum constants where it previously had to spell the raw integer.

- `emit_type` and the value/arith classifiers go through a new shared
  `value_base_type` rail (`daslib/shader_block_layout`), so an enum lands on its
  underlying integer's branch instead of falling off the `Type`-keyed switch.
- `ExprConstEnumeration` resolves to its entry's own initializer, which already
  carries the enum's underlying type — the width/signedness rule is not
  transcribed a second time.
- An enum-typed ternary lowers through `OpSelect`.
- A 64-bit-based enum fails closed, exactly as a bare `int64` does.

Tests: `tests/spirv/test_enum.das`. `tests/spirv` is 347/347.

Note on lint: `spirv_emit.das` and `spirv_types.das` already report 12 STYLE037
and 2 STYLE038 warnings on an untouched `master` checkout (STYLE037/038 landed
2026-07-29 and dasSpirv has not been swept). Nothing here adds one — the enum
lookup lives in its own helper, and the only movement is `emit_const` 76 -> 78,
the same +2 every sibling `?as` case in that dispatch costs.

### DIFF
--- a/daslib/shader_block_layout.das
+++ b/daslib/shader_block_layout.das
@@ -24,6 +24,15 @@ def public matrix_info(t : TypeDecl?) : tuple<ok : bool; width : int; cols : int
     return (ok = false, width = 0, cols = 0)
 }
 
+// The base type a VALUE of this declaration carries: an enum has none of its own and lowers to the
+// OpTypeInt of its underlying integer, which the Type-keyed classifiers below cannot reach. A
+// 64-bit-based enum peels to tInt64/tUInt64, so arith_width_ok refuses it exactly like a bare one.
+def public value_base_type(t : TypeDecl?) : Type {
+    if (t == null) return Type.none
+    if (t.isEnum && t.enumType != null) return t.enumType.baseType
+    return t.baseType
+}
+
 // The whole 16/8-bit type set of the shader arc: the lattice proper (contiguous appended
 // block in the Type enum) PLUS the pre-lattice small scalars, which live in the enum's old
 // region (near tInt) and are NOT covered by the range check.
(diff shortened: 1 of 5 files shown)


olegus8:
The failing `extended_checks (linux, 64)` is the lint step, and all 14
STYLE037/038 warnings it reports reproduce on an untouched `master`:

```
git checkout master
./bin/daslang ./utils/lint/main.das -- --quiet \
  daslib/shader_block_layout.das \
  modules/dasSpirv/spirv/spirv_types.das \
  modules/dasSpirv/spirv/spirv_emit.das
# --- Summary --- 3 files, 14 issue(s), 0 error(s)     (exit 2)
```

Same 14 functions before and after this branch. Per function, the only value
that moves is `emit_const` 76 -> 78 — the +2 that every sibling `?as` case in
that dispatch costs — and the enum lookup itself sits in its own helper
(`emit_const_enum`), which carries no warning. The rest are `visitExprCall` at
272, `classify_global` at 154, `glsl_ext_op` at 61 and so on, none of them
touched here.

STYLE037/038 landed on 2026-07-29 (f0a24511a6, d4ccdfd01f) and dasSpirv has not
been swept since, so any PR touching `spirv_emit.das` fails this step today.

I did not want to change the module's lint policy from inside a feature PR. If
you would rather this went green, say so and I will either add
`options _cyclomatic_complexity = 0` / `options _function_length = 0` to the
dasSpirv files, or send the sweep as its own PR ahead of this one.

PR #3598, opened 2026-07-31
dasSpirv: bind a reference local to the pointer it aliases (stacked on the enum PR)

olegus8:
Stacked on the enum PR — review that one first; this branch's own change is the
last commit.

A reference local aliases storage that already has an address, so it gets no
`OpVariable` of its own: it binds the pointer its initializer lowered to.
Declaring one made every write through the alias land in a private copy and left
the aliased element untouched.

The inliner emits this shape whenever an inlined call's result is assigned to an
lvalue (`var __inl_pre : float& = p[i]; ...; __inl_pre = __inl_res`), which is
how a helper reached across a `require` arrives — so a fixed-array element write
through a required-module function silently dropped.

Tests: `tests/spirv/test_xmod_array_store.das` plus the `_spirv_xmod.das`
helper module. `tests/spirv` is 350/350.

### DIFF
--- a/daslib/shader_block_layout.das
+++ b/daslib/shader_block_layout.das
@@ -24,6 +24,15 @@ def public matrix_info(t : TypeDecl?) : tuple<ok : bool; width : int; cols : int
     return (ok = false, width = 0, cols = 0)
 }
 
+// The base type a VALUE of this declaration carries: an enum has none of its own and lowers to the
+// OpTypeInt of its underlying integer, which the Type-keyed classifiers below cannot reach. A
+// 64-bit-based enum peels to tInt64/tUInt64, so arith_width_ok refuses it exactly like a bare one.
+def public value_base_type(t : TypeDecl?) : Type {
+    if (t == null) return Type.none
+    if (t.isEnum && t.enumType != null) return t.enumType.baseType
+    return t.baseType
+}
+
 // The whole 16/8-bit type set of the shader arc: the lattice proper (contiguous appended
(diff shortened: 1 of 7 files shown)


olegus8:
The failing `extended_checks (linux, 64)` is the lint step, and all 14
STYLE037/038 warnings it reports reproduce on an untouched `master`:

```
git checkout master
./bin/daslang ./utils/lint/main.das -- --quiet \
  daslib/shader_block_layout.das \
  modules/dasSpirv/spirv/spirv_types.das \
  modules/dasSpirv/spirv/spirv_emit.das
# --- Summary --- 3 files, 14 issue(s), 0 error(s)     (exit 2)
```

Same 14 functions before and after this branch. Per function, the only value
that moves is `emit_const` 76 -> 78 — the +2 that every sibling `?as` case in
that dispatch costs — and the enum lookup itself sits in its own helper
(`emit_const_enum`), which carries no warning. The rest are `visitExprCall` at
272, `classify_global` at 154, `glsl_ext_op` at 61 and so on, none of them
touched here.

STYLE037/038 landed on 2026-07-29 (f0a24511a6, d4ccdfd01f) and dasSpirv has not
been swept since, so any PR touching `spirv_emit.das` fails this step today.

I did not want to change the module's lint policy from inside a feature PR. If
you would rather this went green, say so and I will either add
`options _cyclomatic_complexity = 0` / `options _function_length = 0` to the
dasSpirv files, or send the sweep as its own PR ahead of this one.

The list is identical to #3597's — this branch's own commit adds no warning of
its own; `visitExprLetVariable` stays under both ceilings.

PR #3610, opened 2026-08-02
Preview caret: collapsed selection + focus-line tint, not a one-byte span

borisbat:
## Preview caret: collapsed selection + focus-line tint

Live-play catch on E5 (#3607): typing in side-by-side markdown randomly selected whole sections in the preview.

**Root cause**: the caret-follow mark was a **one-byte selection** (`select_source_range(caret, caret + 1)`). When the caret sits on unrendered markdown source — blank lines, emphasis markers, list bullets, i.e. constantly while typing — `source_byte_position` falls forward to the next *rendered* run for each end independently, so anchor and focus land in different runs and the painted selection spans the gap: whole "sections" light up.

**Fix**:
- The caret maps as a **zero-width** selection — position resolution only, nothing paints; the end-of-document back-walk keeps its collapsed form.
- New view helper `markdown_view_focus_line_tint(state, color)` tints the focus **line** across the pane (current-line style, drawn by the view where the layout lives) — the visible cursor indication ruling (e) asked for, without any text selection.
- `text_viewer_state` gains `preview_selection_start/end`; the watch scenario pins the selection staying collapsed after caret moves.
- Two pre-existing PERF030 hits on the view's dispose idiom (reset-after-field-deletes) carry the same reasoned nolints `text_edit_dispose` got in #3607.

Verified: viewer suite 10/10 headless (clean port), lint 0, formatted; driven live — the tint tracks the caret through blank lines and syntax without selections.

### DIFF
--- a/examples/text/main.das
+++ b/examples/text/main.das
@@ -1091,16 +1091,18 @@ def private apply_preview_caret(var doc : TextViewerDocument) {
     return if (doc.preview_revision != doc.edit.revision)
     if (doc.preview_caret_applied != doc.edit.follow_revision) {
         doc.preview_caret_applied = doc.edit.follow_revision
-        // one byte wide (a zero-width selection maps but yields no focus
-        // screen Y), walked back to the nearest RENDERED byte - the caret can
-        // sit on blank lines or fence syntax the preview never draws
+        // ZERO-width: a one-byte selection paints random spans when the caret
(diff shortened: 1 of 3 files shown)


borisbat:
Not landing separately - E5 play-round fixes ride the E5.5 PR (one PR per arc). Branch stays as the accumulator.

Issue #3619, opened 2026-08-04
nightly imgui: text_edit/viewer/hybrid suite red every night — rotating wait_cond timeouts


borisbat:
The nightly imgui lane (`nightly_imgui.yml`) has failed on master every night for at least three nights, with the failing cells **rotating inside the text_edit / text_viewer / text_hybrid family**:

| run | date | failing cells |
|---|---|---|
| 30683361175 | Aug 1 | `test_text_edit_find_replace`, `test_text_viewer_open_modes_and_zoom` |
| 30731934456 | Aug 2 | `test_text_edit_find_replace`, `test_text_viewer_open_modes_and_zoom` |
| 30783823651 | Aug 3 | `test_text_edit_find_replace`, `test_text_edit_keystrokes_reach_buffer`, `test_text_edit_numpad_and_overtype`, `test_text_viewer_edit_mode_save_point`, `test_text_viewer_edit_preview_and_watch` |

A local preflight mirror (macOS, 2026-08-04) drew 3 cells from the Aug-3 set plus `test_create_link_snippet_tab_flow` and `test_hybrid_typing_debounce_and_exit_snap` — same family, same signature.

**Signature:** every failure is a `wait_cond` 30 s timeout, not an assertion — the `log_snapshot` dump shows the app alive and rendering (menus, toolbar, editor widget all present) but the awaited state never observed. `test_text_edit_find_replace` is the constant across all runs; the rest rotate, which reads as timing flake around a real underlying issue rather than N independent bugs.

**Discipline pointer:** per `skills/imgui_ui_debugging.md` / `skills/imgui_playwright.md`, the fix path is gate-on-the-effect — reproduce one cell headless, make the missing state observable in `imgui_snapshot`, and either the harness waits on the wrong condition (fix the wait) or the state genuinely never arrives (real regression — bisect the family, `find_replace` first since it's the constant).

Not caused by any in-flight branch: the failing suite predates and rotates independently of current PRs (verified against a diff touching zero imgui files).

PR #3620, opened 2026-08-04
dasLLAMA: canary lands honest — q8 decoder/encoder, the decode cap, three fresh boards (M1/M4/zen4), fresh-box bringup hardened end to end

borisbat:
The canary-apples arc, squashed. Measurement log: `modules/dasLLAMA/canary_apples_plan.md`.

## Canary-Qwen

- **q8 decoder + encoder flip** with the attention q-block slab work, gated by token-for-token oracle parity and byte-identical transcripts on the fixture clips.
- **Honest decode cap**: `CN_MAX_NEW = 256` mirrors the reference arm's `--max-new-tokens 256`. The runaway class (hp0 292 s vs NeMo 99 s) was the model decoding 4.7k loop tokens against its trained envelope (40 s / 1024 tokens per the model card) — a model limitation we stop forcing, not a perf gap. Spec ladder caps at gb1; void rows dropped from the store.
- Pre-flip zen2 canary rows retired (the quantizer changed under them); zen2 re-measures in its own sweep.

## zen4 — first AVX-512 box, BRINGUP.md executed end to end

Bare Debian 12 to a published 34-cell board in one day. Every seam the bringup caught is fixed here:

- `performance/fetch_models.das` — the committed model-provenance manifest: 18 entries, HF repo + revision pin + sha256, conversion recipes. Parakeet/canary-encoder conversions reproduced **byte-identical cross-arch** (ARM-recorded shas, x86 reproduction); the canary decoder is per-arch by design and gates on transcript parity (passed byte-identical).
- Setup tools stop auto-minting junk tune sidecars: `tune_policy(missing="fallback")` on `setup_asr_rig` and `gen_site_records` (the annotation needs a direct `llvm_tune` require; `setup_lcpp_ref` has no dasllama graph and needed nothing).
- `setup_asr_rig` fetches gb1 itself — the corpus no longer assumes a make-samples-era whisper.cpp checkout.
- `gen_bench_records`: finds the rig exe on bare Linux (`daslang -exe` appends `.exe` on every platform), stamps hardware identity from its own probe (a child exe built before a probe fix carries stale identity — zen4's das cells read "unknown"), and the cpu/os probes get Linux arms (`/proc/cpuinfo`, os-release).
- Box-profile `threads` field records **total lanes** (workers + computing main, the `DAS_JOBQUE_THREADS` unit) and the loader stops advising an env override.
- First AVX-512 mint: paranoid 561.6 s, noise=ok first try, **11 of 30 crowns differ from the per-ISA fallbacks** (wider vectors win: dot→vec16_u2, quantize_q8_0→vec32_u2, rmsnorm→vec16).
- The board (`records/zen4.json`, generation archived): **pp512 2.0–2.8× vs native `-march=native` AVX-512 llama.cpp** across seven of eight models, tg at the DDR5 wall (0.99–1.12×); canary 2.3–3.4× vs NeMo. Output sanity: coherent E4B generation, transcripts byte-identical.

## Site

- **Thread counts explicit in every mode** — a `threads` column in the LLM and audio pair tables plus the receipts (an 8-core box publishes 8, not the policy's 16 ceiling).
- **Annotation-voided clips render no win**: `void_clips` on `BenchAnnotation`/`BenchRun`; the Omni jfk decode-runaway rows (m4/zen2/zen4, up to 73× against a runaway reference) show —✱ with raw times and the note intact. Honest ratios on annotated rows are untouched.
- M1 + M4 boards re-celled same-session; stale fp32-decoder annotation retired.

## Ledgered for the arc remainder

gemma-26B x86 gap (0.50× zen2 → 0.98× zen4, kernel unnamed); audio-tower AVX-512 gap (das tower is iso-speed M1↔zen4 while ggml's x86 tower is ~3× its own ARM one); `fetch_models --convert` arm; zen2 sweep.

## Preflight

17/18 gates green (lint, tests-interp, tests-jit, AOT subset, docs, format all pass). The single red is the imgui gate — master's own nightly has been red on the same rotating text_edit family for 3+ nights (`wait_cond` timeout signature, zero imgui files in this diff): receipts and fix path in #3619, adjudicated as a follow-up.


borisbat:
## Review response — cbd7707d5

Two independent review passes ran against this PR: the standard 5-agent code review, and a CODEREVIEW.md-rules auditor. Their findings were near-disjoint (implementation bugs vs process violations). Disposition of all 19:

**Fixed in cbd7707d5 (13):** the three function-valued globals missing from `restore_math_fn_defaults` (kq-exec in a baked exe would invoke null); per-row 32-alignment guard in `cn_repack`; loud truncation warning when `CN_MAX_NEW` exhausts; `mint_canary_encoder` derives its tag from the staging and is now test-covered (was dead); profiler moved out of `measure_das`'s timed region (canary/whisper rows carried overhead their siblings didn't); `var inscope` across the tuner's bench scratch (~285 MB stayed resident through a mint — verified by a full scratch-manifest mint, all noise probes quiet); `fetch_models` counts only materialized fetches and fails recipe-less absences; `mem_census.sh` survives a failed cell and quotes its args; `das_asr_run` picks newest-by-date matching the site renderer; one shared `list_record_stores` (the two copies had drifted); ASR facade verbs forward to family verbs; `void_clips` trims its parts; CODEREVIEW.md carries the mem_census carve-out + fetch_models rule line.

**Annotated, re-sweep scheduled (1):** the 7 M4 das ASR rows measured under the mis-shipped stale sidecar now carry annotations naming that condition; the M4 re-sweep under the fixed release is a follow-up (hours of box time).

**Declined with rationale (2):** canary's decoder staying q8 under `set_asr_fp32` is the documented design (the facade doc states every decoder loads q8; the fp32 parity arm rides `QuantMode.fp32 + image_fp32` off-facade, and the oracle passes token-for-token on the q8 rail); `rig_exe_path` sharing daspkg's suffix helper would couple two unrelated tools' require graphs for one string — the comment cites the source of truth.

**Deferred with rationale (3):** Parsec/remote-desktop provenance is only honest if stamped at measurement time — it rides the M4 re-sweep; the daspkg `ship_manifest` test cell needs a mint-stub design (a scope-bearing fixture mints for real; the branch is meanwhile verified by this PR's zen4 release runs); `asr_exec_fmt` unification with `model_exec_fmt`'s per-tensor walk is a design question (towers aren't `Model`s).

**Gates re-run after the fixes:** ASR suite 35/35 (`PARITY_FULL`, canary NeMo oracle token-for-token), image arms via the scoped runner (incl. the new mint leg), schema 10/10, box-profile, lint/format clean, site rendered and checked.

borisbat:
## Re-audit residuals — e222c8a0a

The CODEREVIEW.md auditor re-ran after cbd7707d5 and held the bar: 5 residual violations, 3 unproven. All addressed:

- **Provenance reaches every page**: `gen_results` now applies `records/annotations.json` and renders a caveat line under annotated tables — the M4 unstamped-rows note reached the site but not `results.md`. Also found concretely wrong: six M4 sections sat above the m1 `Source:` line and read as m1 data; relocated with their own `Source: records/m4.json` lines.
- **The facade forwards on all six arms**: gemma4a/qwen3a tower verbs added; the catch-alls no longer hardcode "the towers are fp32" in the facade.
- **Moved bits get their cells**: family verbs fed constructed structs, `list_record_stores` strictness over a temp dir, `encode_ms`/`void_clips`/`comment` schema round-trip, annotation targeting (`test_asr_verbs.das` + `test_bench_records_schema.das`).
- **The reference side gets a manifest gate**: the four cli-reference q8 carriers are presence-gated recipe rows in `fetch_models` (verify now 22/22), plus a `models_provenance()` invariants cell; both new files have their ARCHITECTURE §1.8 charters.
- **Unproven → settled**: kq arms green under `PARITY_FULL` (the restore-guard exposure); ASR suite 35/35 re-run after the dispatch change; daspkg unit suite ran 217/217 on macOS during PR prep; the ship-manifest branch was hand-verified on the M4 during the mint saga this PR documents (stated per the rule's "or the review says so" clause).

PR #3641, opened 2026-08-06
stabilize the compiler against null AST pointers

profelis:


src/ast/ast_export.cpp
```diff
@@ -223,8 +223,8 @@ namespace das {
         // function address
         virtual void preVisit(ExprAddr * addr) override {
             Visitor::preVisit(addr);
-            if (builtInDependencies || (addr->func && !addr->func->builtIn)) {
-                DAS_ASSERT(addr->func);
+            if ( !addr->func ) return;
```
aleksisch, line 226: What is the reason these pointers are null? Shouldn't this reason be fixed instead?

Now we simply delay the crash until `simulate` or some later phase?

PR #3720, opened 2026-08-13
dasSpirv: a non-interpolated interface is never Flat

olegus8:
`classify_global` decorates every integer varying `Flat`, because on an
interpolated interface there is no valid interpolation mode for an integer. A
vertex `@in` is not such an interface: it is a vertex attribute the vertex buffer
feeds whole, nothing interpolates it, and SPIR-V forbids the decoration there —
`VUID-StandaloneSpirv-Flat-06202` (Flat / NoPerspective / Sample / Centroid must
not be used on an `Input` variable in a vertex shader). So an int vertex
attribute got a decoration the spec rejects.

The condition gains `!is_attribute` (`is_in && stage == Vertex`) beside the
existing `!is_pervertex`. Every other stage and both the int-kind and explicit
`@flat` paths are untouched — a fragment `@in` still auto-Flats, `@flat` still
honoured on floats, and a vertex `@out` (which IS interpolated) still takes Flat.

`tests/spirv/test_raster.das` gains `test_vertex_attribute_is_never_flat` over a
new `attr_vert` shader in `_spirv_common.das`: an `int` `@in` at Location 3, which
would take Flat under every other interface. It asserts no Flat decoration, that
the Location survives, and validates the module.

Verified locally: `tests/spirv` is 370/370 green with the change, and the new test
fails (1 failed) against the pre-change emitter, so it pins the fix rather than
riding along with it.

### DIFF
--- a/modules/dasSpirv/spirv/spirv_emit.das
+++ b/modules/dasSpirv/spirv/spirv_emit.das
@@ -1117,10 +1117,15 @@ def private classify_global(var m : SpirvModule; var ctx : EmitCtx; v : Variable
         // for those. Floats default to perspective-correct smooth interpolation; honour an explicit
         // @flat annotation when the user wants flat-shaded floats (low-poly stylised look, provoking-
         // vertex picks per primitive). Reflection bit not added — Flat is purely a SPIR-V decoration,
-        // it doesn't change the descriptor or pipeline layout. A @pervertex input is never Flat.
+        // it doesn't change the descriptor or pipeline layout. A @pervertex input is never Flat, and
+        // neither is a VERTEX @in: that one is a vertex attribute the vertex buffer feeds whole, so
+        // an interpolation decoration on it is forbidden rather than mandated
+        // (VUID-StandaloneSpirv-Flat-06202).
         let bt = v._type.baseType
         let is_int_kind = scalar_class(bt) >= 0 && scalar_class(bt) <= 1
-        let want_flat = (is_int_kind || (v.annotation |> find_arg("flat") is tBool)) && !is_pervertex
+        let is_attribute = is_in && stage == ShaderStage.Vertex
+        let want_flat = ((is_int_kind || (v.annotation |> find_arg("flat") is tBool))
+            && !is_pervertex && !is_attribute)
         if (want_flat) {
             emit(m, SEC_DECOR, SpvOp.Decorate, vid, uint(SpvDecoration.Flat))
         }
--- a/modules/dasSpirv/spirv/spirv_emit.das
+++ b/modules/dasSpirv/spirv/spirv_emit.das
@@ -1118,14 +1118,16 @@ def private classify_global(var m : SpirvModule; var ctx : EmitCtx; v : Variable
         // @flat annotation when the user wants flat-shaded floats (low-poly stylised look, provoking-
         // vertex picks per primitive). Reflection bit not added — Flat is purely a SPIR-V decoration,
         // it doesn't change the descriptor or pipeline layout. A @pervertex input is never Flat, and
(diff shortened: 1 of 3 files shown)


borisbat:
Verified the fix end-to-end locally: pre-change, the `attr_vert` shape emits Flat and spirv-val rejects it with `VUID-StandaloneSpirv-Flat-06202`; with the change, Flat count is 0, Location 3 survives, spirv-val clean. The gate is precisely scoped — vertex `@out`, fragment `@in`, and the mesh-output path keep their mandated Flat.

One ask before merge: this fixes half the bug class. The spec forbids interpolation decorations on **two** non-interpolated interfaces, and the mirror one — an int **fragment `@out`** (material-id / picking buffer to an R32I attachment) — still takes Flat with this PR applied. Repro against the patched emitter:

```das
require _spirv_common
require spirv/spirv_shader
require spirv/spirv_dis
require spirv/spirv_grammar

var @out @location = 0 frag_mat_id : int
[fragment_shader(name="probe_frag_spv"), marker(no_coverage)]
def probe_frag {
    frag_mat_id = 7
}

[export]
def main {
    var words <- clone_to_move(probe_frag_spv)
    print("Flat: {count_op_operand_at(words, SpvOp.Decorate, 1, uint(SpvDecoration.Flat))}\n")
    let r = validate_spirv(words)
    print("spirv-val ok={r.ok}\n{r.msg}\n")
    delete words
}
```

```
Flat: 1
spirv-val ok=false
error: line 9: [VUID-StandaloneSpirv-Flat-06201] Flat decorated variable must not be
used in fragment execution model as an Output storage class for Entry Point id 6.
  %3 = OpVariable %_ptr_Output_int Output
```

The suite stays 370/370 only because no fixture has an int fragment `@out` — the fixtures' `mat_id : int` `@out` is on the vertex side.

The extension is the same one-line class as the fix: generalize the gate to both ends of the pipe —

```das
let no_interp = (is_in && stage == ShaderStage.Vertex) || (is_out && stage == ShaderStage.Fragment)
```

— plus a `test_fragment_output_is_never_flat` mirroring the new test. Title then becomes "a non-interpolated interface is never Flat".

olegus8:
Accepted — you're right that it was half the class, and the framing "both ends of
the pipe are not interpolated" is the one that makes the rule obvious. Done in
`69554914f`:

- the gate is now `no_interp = (is_in && stage == Vertex) || (is_out && stage ==
  Fragment)`, exactly as you wrote it, and the comment names both VUIDs;
- `_spirv_common.das` gains `attr_frag` — an `int` `@out` at Location 0 — and
  `test_raster.das` gains `test_fragment_output_is_never_flat` mirroring the
  vertex test: no Flat, Location survives, spirv-val clean;
- retitled to "a non-interpolated interface is never Flat", and the body now
  leads with the two-interface rule.

`tests/spirv` is 372/372 with lint and `dasfmt --verify` clean. Your point about
why the suite stayed green is what the new fixture fixes: every existing `int`
`@out` sits on the vertex side, so nothing in the tree covered a fragment
attachment output until now.

borisbat (approved):
Verified `69554914f` locally, end to end:

- `tests/spirv` 372/372 with the change applied; lint clean on all three files.
- Negative control: with the new tests kept and the emitter reverted to master, exactly the two new tests fail — `attr_vert` emits Flat and spirv-val rejects it with `VUID-StandaloneSpirv-Flat-06202`, `attr_frag` with `-06201`. Both tests pin their half of the fix.
- The generalized gate leaves every mandated Flat in place: `test_vertex_flat` and `test_fragment_flat` (the interpolated interfaces) still pass, and the `no_interp` condition can't reach the mesh-output or compute paths.

The "both ends of the pipe" framing in the comment and PR body is exactly right. Merge-ready from my side once CI reports.

PR #3721, opened 2026-08-13
Diagnostics through one sink

olegus8:
A `TextPrinter` writes to stdout, and so does `DAS_FATAL_LOG` — by its own
`printf`. For a program whose stdout carries a protocol rather than its own
output (a stdio JSON-RPC server: the MCP server, a language server), a leak dump
or a fatal landing there is read as a frame, and the client drops the connection
before anyone sees the text. There was no one place to point that at.

**The sink.** `setTextPrinterSink` is where every `TextPrinter` lands,
process-wide, stdout by default. `textPrinterToStderr()` and
`textPrinterToFile(path)` are the two moves a protocol-carrying program wants;
the file form appends, and returns false without changing the sink when it
cannot open. `DAS_FATAL_LOG` becomes `das_fatal_log`, which writes the same sink
directly — not through a `TextPrinter`, because it fires from the allocator and
the smart-pointer guard, so it may not allocate and may not take the printer's
lock (the fatal it reports can be a crash already holding it).

**What it carries — and what it does not.** Leak dumps (jobque, JobStatus,
smart-ptr tracking), fatals, the JIT's diagnostics, the `exit(N)` notice: the
text the runtime writes *about itself*. **Not** a panic report — that goes out
through `to_err` → `Context::to_out` → the `das_to_stdout_level_prefix_text`
macro, so it never touches `TextPrinter`, and under `logger_install_hook` it
lands in the log like any other level-tagged line. Script `print()` is likewise
untouched.

**The das side.** `diagnostics_to_stderr()` / `diagnostics_to_file(path)` bind
the two sink moves, and `logger_capture_diagnostics()` in `daslib/logger` names
the file for you — `<log>.diag.log` beside the log, stderr if it cannot be
opened. Raw text, not ndjson, and no hook sees it: the runtime prints it, not
`print`. `serve_stdio` calls it before the first frame, which is the point of
the change.

`doc/reflections/das2rst.das` gains the two builtins and
`logger_capture_diagnostics` in existing groups, and both new builtins carry a
handmade description, so the doc job stays quiet.

**Tests.** `tests-cpp/small/test_text_printer_sink.cpp` installs a capturing sink
and pins three branches: `TextPrinter` routing, `das_fatal_log` formatting, and
the fail-closed `textPrinterToFile` (returns false, sink unchanged — proved by
writing again afterwards). `tests/fio/diagnostics_sink.das` pins the das binding
both ways. Default behaviour is unchanged everywhere: the sink starts at stdout,
and a program that never calls these keeps exactly the output it has now.

---

**The lint lane is red on `utils/mcp/mcp_core.das`, and this branch did not cause
it.** `extended_checks` lints every changed `.das` whole, and that file already
carries `PERF030` (the `extra_list` move) and `STYLE037` (complexity 22 on
`handle_tools_call`) at master's tip. The file was last touched 2026-07-20;
`STYLE037`'s limit dropped to 20 on 07-29 and `PERF030` landed on 08-01, so both
postdate it. This branch's own hunk there is the single
`logger_capture_diagnostics()` call and introduces neither. Say the word and I
will fix either here: `PERF030` is a one-line fold of the move into its
declaration (probed: lint-clean, behaviour identical), and `STYLE037` on that
flat argument-extraction shape is what `skills/style_lint.md` says to suppress
with a reason rather than split.

### DIFF
--- a/daslib/logger.das
+++ b/daslib/logger.das
@@ -39,9 +39,9 @@ module logger shared public
 //! What reaches the file is still the level filter's to decide, and
 //! ``print`` is a debug-level line: under the ``LOG_INFO`` default it is
 //! dropped rather than recorded. One install covers every Context including
-//! future ``new_thread`` workers. The runtime's own crash text is not
-//! ``print`` and no hook sees it: :func:`logger_capture_diagnostics` is
-//! what puts that beside the log.
+//! future ``new_thread`` workers. A leak dump or a fatal is not ``print``
+//! and no hook sees it: :func:`logger_capture_diagnostics` is what puts
+//! those beside the log.
 
 require daslib/json public
 require daslib/fio public
@@ -313,9 +313,9 @@ def logger_install_hook() {
 }
 
 def logger_capture_diagnostics() {
-    //! Send the runtime's own diagnostics — a panic report, a leak dump, a fatal — to
-    //! ``<log>.diag.log`` beside the log file, or to stderr when it cannot be opened.
-    //! Raw text, not ndjson, and no hook sees it: the runtime prints it, not ``print``.
+    //! Send the runtime's own diagnostics — a leak dump, a fatal, what it reports
+    //! about itself — as raw text to ``<log>.diag.log``, or to stderr when that
+    //! cannot open. Not a panic report: that is level-tagged, so the hook logs it.
     if (empty(log_path) ||
         !diagnostics_to_file(replace_extension(log_path, ".diag.log"))) {
         diagnostics_to_stderr()
--- a/daslib/logger.das
+++ b/daslib/logger.das
@@ -33,11 +33,15 @@ module logger shared public
 //! stays at the global :func:`logger_set_min_level`.
 //!
 //! :func:`logger_install_hook` plugs a GLOBAL debug agent into the runtime
-//! so accidental ``print()`` / ``to_log()`` calls are diverted into the
-//! log file instead of stdout/stderr — important for stdio-transport
-//! processes (MCP server, language servers) where any stdout write
-//! corrupts the wire format. One install covers every Context including
-//! future ``new_thread`` workers.
+//! so accidental ``print()`` / ``to_log()`` calls are diverted off
+//! stdout/stderr — important for stdio-transport processes (MCP server,
+//! language servers) where any stdout write corrupts the wire format.
+//! What reaches the file is still the level filter's to decide, and
+//! ``print`` is a debug-level line: under the ``LOG_INFO`` default it is
+//! dropped rather than recorded. One install covers every Context including
+//! future ``new_thread`` workers. The runtime's own crash text is not
+//! ``print`` and no hook sees it: :func:`logger_capture_diagnostics` is
+//! what puts that beside the log.
 
 require daslib/json public
 require daslib/fio public
@@ -306,6 +310,16 @@ def logger_install_hook() {
     //! Install a global debug agent that diverts ``print`` and ``to_log`` (from any Context, current or future) into the configured log file. Idempotent — safe to call from every module's [init], from MCP main, etc. Worker threads spawned via ``new_thread`` inherit the hook automatically; no per-context re-install needed.
     return if (has_debug_agent_context(LOGGER_AGENT_CATEGORY))
     install_new_debug_agent(new LogHookAgent(), LOGGER_AGENT_CATEGORY)
+}
+
+def logger_capture_diagnostics() {
+    //! Send the runtime's own diagnostics — a panic report, a leak dump, a fatal — to
+    //! ``<log>.diag.log`` beside the log file, or to stderr when it cannot be opened.
+    //! Raw text, not ndjson, and no hook sees it: the runtime prints it, not ``print``.
+    if (empty(log_path) ||
+        !diagnostics_to_file(replace_extension(log_path, ".diag.log"))) {
+        diagnostics_to_stderr()
+    }
 }
 
 def logger_init(name : string) {
--- a/doc/reflections/das2rst.das
+++ b/doc/reflections/das2rst.das
@@ -193,7 +193,7 @@ get_value|insert_clone|emplace_new|insert_default|emplace_default|get_with_defau
         group_by_regex("Smart ptr infrastructure", mod, %regex~(add_ptr_ref|smart_ptr|get_const_ptr|move|move_new|get_ptr$)%%),
         group_by_regex("Macro infrastructure", mod, %regex~(is_folding|is_compiling|is_in_completion|is_in_lint_check|is_building_documentation|is_reporting_compilation_errors)%%),
         group_by_regex("Profiler", mod, %regex~(profile|reset_profiler|dump_profile_info|collect_profile_info)$%%),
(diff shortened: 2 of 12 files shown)


borisbat:
Verified the mechanism against the tree, and most of the PR's claims hold:

- The `TextPrinter` census confirms the sink captures exactly the runtime's self-diagnostics — jobque leak dumps, JobStatus/smart-ptr tracking dumps, the ABI-stamp fatals, the `exit(N)` notice, JIT diag. Script `print()` does not flow through it (it goes via `Context::to_out`), so `diagnostics_to_*` doesn't overreach its name.
- `das_fatal_log`: decl/def both global-scope, no-alloc/no-lock honoured, `#ifndef DAS_FATAL_LOG` embedder override preserved.
- `textPrinterToFile`: file swapped in before the sink, append mode, fail-closed `false`, deliberate never-close — all sound under the stated set-before-threads-print contract.
- The lint preemption is exactly accurate: `mcp_core.das` at master's tip carries precisely PERF030@459 and STYLE037 (complexity 22)@417 without this branch.
- The rewritten logger docstring's "`print` is a debug-level line" is true (`defaultPrint = debug`, `string_writer.h:138`).

Two asks before merge:

**1. The "panic report" claim is wrong in all five doc sites, and it cuts against the PR's own motivation.** An unhandled panic prints via `to_err` → `Context::to_out` → the `das_to_stdout_level_prefix_text` macro (`das_config.h:152`), which sends error-level text straight to **stderr** — or the logger hook eats it into the main log. It never touches `TextPrinter` or `DAS_FATAL_LOG`, so `diagnostics_to_file` will never put a panic report in `<log>.diag.log`. Yet the PR body, the `logger_capture_diagnostics` docstring, both handmade RSTs, and the `serve_stdio` comment all lead with "a panic report". For the stated scenario — a launcher that records neither stream — the most common death rattle still lands on unrecorded stderr. In the MCP server nothing is lost in practice (the hook records the panic in the main log), but someone debugging a dead server will look in `.diag.log` for it and it won't be there.

Fix in this PR: say what the sink actually carries (leak dumps, fatal reports, runtime self-reports — panic reports reach the main log via the hook). Routing the `to_out` fallthrough through the sink too is a separate discussion — that macro is itself an embedder override point.

**2. No tests.** Every new branch is unpinned. Cheap set: a C++ doctest with a capturing sink pins `TextPrinter` routing and `das_fatal_log` formatting; a bad-path `textPrinterToFile` call pins the fail-closed branch (returns false, sink unchanged); a das test pins `diagnostics_to_file` true/false.

Nit: both handmade RSTs say "the runtime own diagnostics" — missing the "'s".

olegus8:
Both accepted. `65f6c2ffa`.

**1. The panic claim is gone, and you're right that it cut against the PR's own
motivation.** A panic goes out through `to_err` → `Context::to_out` → the
`das_to_stdout_level_prefix_text` macro, so it never reaches `TextPrinter` or
`DAS_FATAL_LOG` and `diagnostics_to_file` will never put one in `.diag.log`.
Every site now says what the sink actually carries — a leak dump, a fatal, what
the runtime reports about itself — and the `logger_capture_diagnostics` docstring
plus both RSTs say outright that a panic report is *not* among them and reaches
the log through the hook instead. I also fixed the same wrong claim in two places
you didn't list: the `setTextPrinterSink` comment in `string_writer.h` and the
`diagnosticsToStderr` comment in `module_builtin_runtime.cpp`. The missing "'s"
is fixed in both RSTs.

Routing the `to_out` fallthrough through the sink as well is, as you say, a
separate discussion — that macro is an embedder override point, and changing it
would move panic text for every embedder, not just protocol-carrying ones.

**2. Tests.** `tests-cpp/small/test_text_printer_sink.cpp` installs a capturing
sink and pins three branches: a `TextPrinter` routes through it rather than
stdout, `das_fatal_log` formats through the same sink (`"fatal %s %d\n"` →
`"fatal here 7\n"`), and an unopenable `textPrinterToFile` returns false *and*
leaves the sink where it was — the fail-closed branch proved by writing again
afterwards and finding the text still in the capture. A guard restores the
default sink per subcase, since it is process-wide.

`tests/fio/diagnostics_sink.das` pins the das side both ways: false for a path
that cannot be opened, true for a temp path. It then points the sink at stderr
before removing the temp directory, because the sink outlives the test and never
closes the file it holds.

The das test is green locally (3/3, `dasfmt --verify` and lint clean). The C++
one I can't build here — this box has no daScript build — so its verdict is CI's;
it uses the same include set as the existing small tests that already construct a
`TextPrinter`.

borisbat:
Ran the verification you couldn't: built `65f6c2ffa` merged with master on MSVC and gave the C++ test its first run — 1 case, 4 assertions, all pass, and each subcase behaves as designed (capture routes, fatal formats, the unopenable file leaves the sink captured). `tests/fio/diagnostics_sink.das` is 3/3 here too. On CI every build lane is green, including the three sanitizer lanes that run the suite under ctest.

**The lint lane is settled on our side.** Confirmed from the job log it fails on exactly the pre-existing pair you diagnosed (PERF030:459, STYLE037:417) and nothing new — your own four changed `.das` files lint clean. That debt predates you and is ours: #3725 retires both (your probed PERF030 fold, and the documented STYLE037 suppression). Nothing to do on this branch; the lane goes green on re-run once that merges.

**One ask, found by running your das test on Windows: the sink's file handle outlives the cleanup.** `diagnostics_to_stderr()` flips the sink pointer but `g_textPrinterFile` keeps the open `FILE*`, and `fopen` on Windows doesn't share delete — so the test's `rmdir_rec(root)` fails silently and every run leaks a `das_diag_*` directory in `%TEMP%` (my run left `das_diag__960281815` behind). The comment in `textPrinterToFile` already states the contract that makes the fix safe: nothing prints while the sink is being set, which is exactly when closing is legal. Close `g_textPrinterFile` when the sink moves *off* the file — in `setTextPrinterSink`, or in `textPrinterToStderr` — so "never closed" narrows to "never closed while it is the sink" (a fatal during shutdown still lands wherever the sink points). Then the test can assert `rmdir_rec` succeeded, which pins the close.

olegus8:
Thanks for building and running it — and for the leaked `das_diag__960281815`,
which is the part I could not have found here: on Linux an open handle does not
hold a file against deletion, so `rmdir_rec` succeeds either way and my test was
green on a bug.

Fixed in `7541dd192`. `setTextPrinterSink` now closes `g_textPrinterFile` when the
new sink is not the file sink, so "never closed" narrows exactly as you put it —
never closed *while it is the sink*. `textPrinterToFile` keeps its own close of
the handle it replaces, since there the sink does not change kind and setting it
closes nothing; the comment now says which of the two closes is which. A fatal
during shutdown still lands wherever the sink points, because nothing closes the
file until something moves the sink off it.

`tests/fio/diagnostics_sink.das` asserts the removal rather than ignoring it:

```das
diagnostics_to_stderr()
t |> success(rmdir_rec(root), "the sink held its file open")
```

That assertion is a Windows-side pin by nature — it cannot fail on this box for
the reason above — so its verdict is your MSVC run and CI's Windows lanes, not
mine. Locally it is 3/3 with lint and format clean.

On the lint lane and #3725: understood, nothing from me on this branch. Happy to
rebase onto it once it lands if you would rather see the lane green here before
merging, but as it stands the four changed `.das` files lint clean and the pair
is yours.

PR #3722, opened 2026-08-13
Fold a demanded const even where folding is off, and expose the question to das

olegus8:
A lint compile has folding off, so a const's initializer arrives whole: a const
global spelled `2 + 2` or as another const's name, a `float4(…) / 2`. Anything
that needs the VALUE — a backend reading a global, a macro asking what a constant
is worth — then has nothing to read, and hand-rolling an is-this-constant test at
each site gets the easy shapes and misses the rest.

**The fold on demand.** `foldConstInit` moves out of `ast_infer_type_helper.cpp`
down into `FoldingVisitor` (`ast_const_folding.cpp`), which is where the folding
machinery already lives — most of the `-123 / +159` line count is that
relocation, not new logic. It runs where the const is READ rather than in the
module that declares it, so a const named across a module boundary resolves, and
it is confined to what a constant was actually demanded of.

**What is new for a dimension is a CALL.** An operator chain in a dimension
(`int[ROWS + 1]`, and a const-of-const, across modules too) already folded at the
dim site before this change. `int[int(4.9) + 1]` did not — it was
`error[30109] array dimension must be constant` — and that is what the `ExprCall`
arm plus the `TypeDecl` force close.

**The same question from das.** `get_const_expr(program, expr)` returns a fresh
constant expression or `null`. It records no access-fold on the globals it reads,
so asking never suppresses the unused-variable or dead-store lint on them — that
is what makes it safe to call from a lint-time macro. The answer lands on
whichever gc root is in force, so application code wraps the ask in
`ast_gc_guard()`; `skills/das_macros.md` says so.

**Each const folds once per ask.** A const read N times used to cost its whole
init N times over, so a chain whose every level reads the one above it was
exponential in the chain's length — 6.5 s at 12 levels against a 0.26 s floor,
and hours a few levels later. `getConstExpr` now memoises per `Variable`, which
makes it linear: 80 levels cost 0.25 s, indistinguishable from the floor. A memo
hit records `access_fold` exactly as a miss does and always returns a fresh
clone. Cycles are still broken by the in-flight set, which is what leaves the
standard init-loop diagnostic to report them.

**Tests.** `tests/lint/test_dim_expr_folding.das` drives a fixture through a lint
compile and pins each shape: a call in a dimension at both ends of a nested array
(the shape that fails without this change), a dimension expression at either end,
a const whose initializer is arithmetic, one read across a module, a constructor
call, and a 40-term sum. It also asserts the negative — that a `PERF020` site in
the same fixture is still reported after the folding ran, i.e. asking did not
fold the lint's own evidence away.

That last assertion is why `_dim_expr_folding_fixture.das` deliberately keeps one
unsuppressed `PERF020` (`let UNDEMANDED = int(5)`), and the changed-file lint lane
reports it. Suppressing it would delete the instrument the test measures with;
`tests/lint/_nolint_lint003_fixture.das` carries an unsuppressed `LINT003` for the
same reason today.

Locally: `tests/lint` is 44/44, the full suite is 10364/10421 with every failure
tracing to a prerequisite this tree does not build, all changed `.das` pass
`dasfmt --verify` and `--ast-verify`. The C++ verdict is CI's.


borisbat:
Verified the mechanics: the relocation into `FoldingVisitor` is faithful, normal compiles keep the old behaviour (`demandFoldConstInit = false` makes `foldConstInit` ≡ `getConstExpr`), the `recordConstAccess` split is a good instrument and the ROWS-recorded / STEP-unrecorded pair pins it well, and the deliberate `PERF020` in the fixture has the `_nolint_lint003_fixture.das` precedent. Two findings, one measured.

**1. The dim half has no negative control — every dim shape in the fixture already compiles at master.** I compiled the fixture's dim shapes (outer expr `int[ROWS + 1][2]`, inner expr `int[2][ROWS * 2]`, const-of-const `int[SLOTS]`, and the cross-module `int[LANES + 1]`) on a pre-change binary under the test's exact policy set (`export_all + lint_check + no_optimizations + no_infer_time_folding`): `ok=true`. The old `getConstExpr` at the dim site (`ast_infer_type_helper.cpp:723`) already folds Var/Op1/Op2 chains, across modules too — so `test_dimension_expression_folds_without_infer_time_folding` passes without this PR's C++. What actually fails at master is a **builtin call in a dimension**:

```das
options infer_time_folding = false
struct D {
    x : int[int(4.9) + 1]    // error[30109]: array dimension must be constant
}
```

That is the shape the new `ExprCall` arm + the `TypeDecl` force genuinely close. Put it in the fixture — at both ends of a nested array, so the one-bool force/restore discipline is pinned too — and the first test earns its name.

**2. The "32 consts read" bound exists only in prose, and its absence is measurable.** The PR body and the `skills/das_macros.md` hunk both state it ("gives up only past 32 consts read to answer one"); the code's only guard is the in-flight cycle set, which breaks cycles but bounds no work — there is no memoization, so every read of a const re-folds its whole init subtree. Measured at master (same recursion, reachable through a dim) with a value-preserving 3-branch chain, folding off:

```das
let A12 = 1
let A11 = A12 + A12 - A12   // = 1, three reads per level
...
let A0 = A1 + A1 - A1
struct DD { x : int[A0 + 1] }
```

n=8: 281 ms, n=10: 535 ms, n=12: 2774 ms — ×9 per two levels, so ~20 lines of legal code puts a lint compile into the hours. Pre-existing, but this PR promises the bound in two documents and newly hands the same recursion to every das macro via `get_const_expr`. Implement the counter the docs already describe (consts read per outer ask, give up past the cap) — or memoize per variable, which kills the blowup outright — rather than editing the docs down to match the code.

Smaller:
- While probing the cross-module dim at master I hit a false `STYLE030`: a `require` whose only use is a dimension expression reports "require unused". Worth checking whether this branch's in-place dim fold changes that answer in either direction — adjacent lint bug either way.
- `isConstExprFunc` now exists twice: the new static in `ast_const_folding.cpp` (null-guards `fun->result`) and the retained `InferTypes::isConstExprFunc` (doesn't). One predicate, one definition.

olegus8:
Both findings accepted and fixed in `c02ed5098`; the measured one is measured
back.

**1. The dim half had no negative control — correct, and the fixture now carries
one.** `Called` holds a builtin call in a dimension at both ends of a nested
array:

```das
struct Called {
    outer : int[int(4.9) + 1][2]
    inner : int[2][int(4.9) + 1]
}
```

The inner one is what pins the force/restore discipline, since it folds under the
same bool the outer site set. `tests/lint` stays 44/44. I have also stopped the
PR body claiming the operator shapes needed this change — as you measured, the
dim-site fold already reached `Var`/`Op1`/`Op2` chains across modules, so what is
new for dimensions is the call arm, and what is new elsewhere is the same answer
being available to das.

**2. The bound was prose, the blowup was real, and memoising kills it.** Your
repro on our build, folding off, `n` = chain length, whole-process wall time:

| n | before | after |
|---|---|---|
| 8 | 0.90 s | 0.29 s |
| 10 | 0.94 s | 0.27 s |
| 12 | 6.49 s | 0.26 s |
| 20 | — | 0.34 s |
| 40 | — | 0.28 s |
| 80 | — | 0.25 s |

The compiler floor on this box is ~0.26 s, so after the change the fold is not
measurable at any depth I can write, and n=80 costs what n=0 costs. The values
are unchanged (`int[A0 + 1]` still yields 2 at every n). `getConstExpr` now
memoises per `Variable` on the visitor: a hit records `access_fold` exactly as a
miss does, and always hands back a fresh clone, since a node has one parent.
`demandFoldConstInit` is set once in the `InferTypes` constructor and never
toggled, so a memo cannot cross between demand modes.

**I did not implement the 32-count, and I would rather not.** With memoisation
the work is linear in the number of distinct consts, so a counter buys no
protection — and a hard cap is a behaviour regression: a chain of 40 consts
defined off each other compiles today (0.28 s above) and a cap of 32 would start
refusing it. So the docs now describe what actually bounds the work — the
in-flight set breaks cycles, and each const folds at most once per ask — rather
than a cap that never existed. If you would rather have the counter anyway, say
so and I will add it with a cap high enough to be unreachable by legal code.

**`isConstExprFunc`: one definition.** It is now a free function in `namespace
das` declared beside `FoldingVisitor`; `InferTypes::isConstExprFunc` delegates to
it, so the three infer call sites gain the `fun->result` null guard they lacked.

**The `STYLE030` note: reproduced, and this branch moves it in neither
direction.** A module whose only use is a dimension expression still reports
"require unused" with this branch applied — minimal case is a `require` plus
`across : int[DIM_LANES + 1]` and nothing else. Our own fixture escapes it only
because it also reads `LANES` in a non-dimension const. So it is an independent
lint bug; happy to take it in its own PR if you want it.

Full suite on this build: 10364 / 10421 passed. Every failure traces to a
prerequisite this tree does not build — 51 `error[20605]` for missing shared
modules, `reflection.das` because its inline sample does `require UnitTest`, plus
the LSP host and the JIT dll-cache tests. None are in the fold path, and CI is
the arbiter for that.

borisbat:
Verified the revision end to end on MSVC (PR merged with master):

- **Negative control now earns its name**: the exact `Called` shape — `int[int(4.9) + 1]` at both ends of the nesting — dies on a pre-change binary with `error[30109]` at both sites, and `tests/lint` is 44/44 with the change. 
- **The memoisation holds on my box too**: the n=12 three-branch chain that cost 2774 ms at master compiles in 211 ms with the PR — process floor. And accepted on the counter: you're right that a cap defends nothing memoisation didn't already kill, and it would start refusing a legal 40-const chain. Docs describing the in-flight set + fold-once-per-ask is the honest contract.
- **Probed the force/restore past the fixture's shapes**: a dim expression carrying a fully-visited nested TypeDecl beside the call arm (`int[typeinfo sizeof(type<int[4]>) - 15 + int(4.9) - 4]`, plus a two-dim variant) folds correctly — the one-bool discipline survives a nested type visit inside the dim, which was the corner I most suspected.
- `isConstExprFunc` unification checked: the delegation gives the three infer sites the null guard they lacked. The `STYLE030` dim-use bug is ours — we'll take it, no PR needed from you.

**One ask, and the red lane disappears with it: rename the fixture to the prefix the lint runner already skips.** `utils/lint/main.das` has always skipped `_fixture_*.das` as "(lint rule fixture)" — deliberate-findings files are a solved problem, re-enabled via `--lint-fixtures` for e2e runs. Your file spells the convention backwards (`_dim_expr_folding_fixture.das`), so the changed-file lane lints it and reds on the deliberate PERF020. Rename to `_fixture_dim_expr_folding.das` plus the one `fixture_path()` string; the in-process `compile_file` + `perf_lint_collect` instrument is untouched by the runner's skip, so the test keeps exactly what it measures. Your precedent observation was righter than you knew — `_nolint_lint003_fixture.das` is misnamed the same way and has been redding the lane on every touch; #3725 renames it on our side. (`_dim_expr_folding_consts.das` is lint-clean and needs no rename.)

With that rename this is merge-ready from my side.

olegus8:
Renamed in `d1570cf98`: `_dim_expr_folding_fixture.das` ->
`_fixture_dim_expr_folding.das`, plus the one `fixture_path()` string. I had read
`is_skip_file` and stopped there, and missed the `_fixture_` skip in `lint_file`
one level down — so I argued for the precedent when the convention already solved
it. Thank you for the correction.

The lane's own command now exits 0 on the changed set (3 files, 0 issues, 1
skipped as "(lint rule fixture)"), and `tests/lint` is still 44/44 — the
in-process `compile_file` + `perf_lint_collect` instrument does not go through
the runner's skip, exactly as you said. `_dim_expr_folding_consts.das` keeps its
name.

Good to hear the nested-TypeDecl probe holds: a dim carrying a fully-visited
nested type beside the call arm was the corner I could not reach without a build,
and it is the one that would have made the single restore bool wrong.

PR #3727, opened 2026-08-14
ast_verify: guess a location for a node whose `at` is unset

aleksisch:
## Problem

`--ast-verify` reports a node whose `at` is unset — the invariant every macro-built node is supposed to carry. The report itself had nowhere to point:

- with an enclosing function, it anchored to the function's **head**, so the caret sat on `def foo`, not on the offending construct;
- with no enclosing function (a global initializer, a struct field type, a TypeDecl in a signature) it printed **no file and no line at all**, plus a dangling `" at "` in the log line.

You got a node kind and nothing to navigate to.

## What changed

The location is now guessed from the tree around the offender, in three tiers, each naming itself in the message:

1. the first located node **inside its own subtree**
2. the last located node the walk **passed**
3. the last located **non-expression site** (argument, field, global, function)

Tier 1 runs on demand: when a node turns out to have no `at`, a throwaway visitor scans its subtree for the first node that has one, so a well formed tree pays nothing. Tier 2 needs no ancestor chain — pre-order means the node the walk passed last *is* the enclosing node whenever the offender is a first child, which is why adding an explicit "enclosing" tier measured *worse* (see the table). Tier 3 covers a TypeDecl verified outside any function or expression, where the first two have nothing.

The guess is stamped into the node, which is also what keeps `ExprConst` — pre-visited twice for one node, base then derived — from reporting itself twice.

The infer-time verifier gets the same predecessor fallback, and neither verifier prints a dangling `" at "` when nothing can be found.

Same fixture, three generations:

```
def lost_one(a, b : int) : int {   <- before: function head
    return a + b                   <- interim: last located node, on `return`
           ^                       <- now: 10:11, caret on the operand
```

```
AST verify (post-infer): ExprOp2 has no source location (at is unset) in 'lost_one';
    first located node inside it is ExprRef2Value at at_unset.das:10:11
```

## Why this order — measured, not guessed

Temporary instrumentation in the post-infer verifier: for every node that *has* an `at`, record its true location plus what each candidate predictor would have said at that moment, then score `|predicted_line - true_line|`. A predictor that produces nothing is charged a 60-line penalty; a cross-file answer counts as nothing. Corpus: 43 files (`tests/language` sample + `dastest`, `utils/detect-dupe`, and `daslib` pulled in through requires) — **80,374 located expressions, 16,132 TypeDecls** after dedupe.

| anchor | mean lines off | exact line | exact caret | no location |
|---|---|---|---|---|
| **first located node inside subtree -> predecessor** *(shipped)* | **2.18** | **91.3%** | **46.8%** | 0.2% |
| descendant -> enclosing -> predecessor -> fn | 2.46 | 91.2% | 46.9% | 0.1% |
| enclosing (parent) -> descendant | 4.98 | 80.1% | 30.0% | 0.4% |
| last located node only | 5.78 | 78.4% | 26.1% | 3.9% |
| enclosing only | 6.99 | 78.1% | 28.5% | 4.4% |
| function head *(previous behaviour)* | 28.28 | 7.7% | 3.9% | 1.9% |
| *per-node oracle (upper bound)* | *1.46* | *94.9%* | - | - |
| *per-kind optimal table, 76 kinds* | *1.42* | - | - | - |

Reading the table:

- **descendant beats predecessor and parent** because a node begins where its subtree begins. `ExprBlock` is the extreme case: 1.02 mean vs 25.60 for last-located.
- **an explicit enclosing tier is not worth having**: the shipped two-node chain scores better than the four-tier one, because in pre-order the predecessor already *is* the enclosing node for a first child. The last row of the table is what the shipped chain leaves on the table against a per-node oracle: ~0.7 lines. Tier 3 removes the residual 0.2% that produce no location at all.
- **a per-kind table is not worth it** — 76 hardcoded entries buy 1 line of mean error over the plain chain, and rot on every new node kind. The chain already sits within ~1 line of the per-node oracle.
- **TypeDecls are different, and the proxy misleads there**: a type's own `at` is its *declaration* site, often in another module, while the actionable place is the use site. Confirmed against the population of nodes that genuinely lack `at` today: the descendant (`firstType`) is essentially never located, enclosing/predecessor ~40%, function/argument always. Types therefore get no descendant tier.

## Cost — cheaper than master on real work

`at` is nearly always there, so the per-node path has to do nothing when it is: a null test, one pointer store, and a `need_typed` whose label is built only if it reports. Master interpolated `"{expr.__rtti}"` **twice per node** and threw both away for every node that was fine. The subtree search runs only for a node that turns out to have no `at`.

Measured with callgrind against a pinned binary (wall clock and CPU time were unusable — a concurrent build kept relinking the binary mid-run):

| input | branch | master | delta |
|---|---|---|---|
| `utils/ast-fuzz/selftest/cli_victim.das` (trivial file) | 1901M | 1842M | +58.1M (+3.16%) |
| `utils/detect-dupe/main.das` (large program) | 43484M | 44581M | **−1096.8M (−2.46%)** |

The extra is **fixed, not per-node**: compiling a bigger `ast_verify` module (the `Anchor` struct, the helpers, the `FirstLocatedVisitor` class), paid once per run and only when the verifier is loaded at all. On a real program that same +58M is swallowed by the ~1.15G the walk saves.

For the record, the first version of this PR kept a per-expression frame stack and a `visitExpression` hook to reach ancestors: **13% over master** in wall clock, for 0.28 lines of *worse* mean error. It is gone.

### DIFF
--- a/daslib/ast_verify.das
+++ b/daslib/ast_verify.das
@@ -182,7 +182,9 @@ class private AstVerifyVisitor : AstVisitor {
 
     def report(at : LineInfo; msg : string) : void {
         error_count++
-        to_log(LOG_ERROR, "AST verify: {msg} at {describe(at)}\n")
+        // a detached synthetic tree has no location in it at all - do not trail an empty " at "
+        let loc = describe(at)
+        to_log(LOG_ERROR, empty(loc) ? "AST verify: {msg}\n" : "AST verify: {msg} at {loc}\n")
         compiling_program() |> macro_error(at, "AST verify: {msg}")
     }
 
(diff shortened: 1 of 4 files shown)


borisbat (approved):
Verified on Windows/MSVC. All five fixture pins reproduce by direct invocation (`--ast-verify` on `at_unset.das` — the dastest ast-fuzz suite itself can't spawn under this box's multi-config layout, a known local issue unrelated to this PR): the subtree tier lands the caret on the operand, the predecessor tier on the enclosing `return`, the site tier on argument then function head, and the search skips the located-but-never-run `assume` spelling exactly as the verify walk does.

Checked the design claims against the source: the hot path really is a null test plus one pointer store where master interpolated two strings per clean node and discarded them; the search visitor's two gates are exactly the post-infer verifier's two subtree gates (the third `canVisitQuote` gate in the file belongs to the gc-reach verifier); `track = false` on types keeps a cross-module declaration site from poisoning the anchor; and stamping the guess follows the existing stamp-the-function-head precedent, including the ExprConst double-previsit dedupe.

Cross-validated the cost story here: wall clock on a large program (`--ast-verify` compile of detect-dupe), three runs each way — this branch 11.23–11.27 s, master 11.3–12.8 s and never faster. Same direction as your callgrind numbers, on a different compiler and OS.

Thanks — the measured predictor table, and killing the 13%-slower first version on data, is exactly how a change like this should be made.

PR #3729, opened 2026-08-14
jobque: join the capture worker before its channel scope closes

olegus8:
`run_with_progress` runs the blocking capture on a worker and polls the progress
file on the main thread, taking the worker's result with
`pop_with_timeout_clone`. That pop is not a fence: the worker pushes and
*then* releases its channel reference, and the push is what wakes the consumer —
so the pop can return while the worker's reference is still out. The
`with_channel` scope then closes on a channel still in use, which panics
(`synch primitive deleted while being used`) and takes the process down mid-call,
followed by a fatal from libc when the worker touches the destroyed mutex.

`ch |> join()` before the scope ends is the fence: it waits for the count to
reach zero, which is exactly what the pop does not do.

The old comment said the pop was the whole teardown because
`notify_and_release` had already balanced the worker's reference. That is true of
the *count*, and beside the point: the release happens after the push, so the two
are not ordered against the consumer.

`skills/jobque_debugging.md` gains what that failure looks like from the dump —
a leak report following an `EXCEPTION:` line is a consequence rather than the
leak, and `ref=N` on a `with_channel` line is the scope closing on a worker that
has not released. It also names which consumers are already fenced
(`for_each_clone`, whose final `pop` blocks on the count) so the rule reads as a
distinction rather than "always join".

Locally: `mcp_core.das` lints clean and passes `dasfmt --verify`.
`utils/mcp/test_tools.das` reports 423 tests / 397 passed / 26 failed on this
box — and **the same 26 fail on untouched master**, so the run carries no delta
from this branch; the failures are environmental here (the tool suite drives
external binaries this box lacks).

### DIFF
--- a/skills/jobque_debugging.md
+++ b/skills/jobque_debugging.md
@@ -73,6 +73,22 @@ total 1 leaked Feature objects
 - **created at** — daScript source location
 - **owner=#4** — Feature is inside JobStatus #4
 
+**A dump that follows an `EXCEPTION:` line is a consequence, not the leak.**
+A panic is fatal to the process, so every reference live at that moment is
+reported as leaked whatever the panic was. Chase the exception, not the
+refcounts.
+
+**`synch primitive deleted while being used (ref=N)` on a `with_channel` line
+is that scope closing on a worker that has not released yet** — N is the count
+it still holds, and the worker then touches a destroyed mutex, so the panic is
+followed by a fatal from libc. Taking the worker's payload is not that release:
+a worker pushes and releases after, which wakes the consumer between the two,
+so a consumer that stops at the payload (`pop_with_timeout_clone`, `try_pop`)
+must `join()` the channel before its scope ends. A consumer that stops on the
+count instead — `for_each_clone`, whose final `pop` blocks until the count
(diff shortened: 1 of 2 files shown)


borisbat (approved):
Reproduced the race deterministically on Windows: the pop-then-scope-close shape with a 200ms worker delay between `push_clone` and `notify_and_release` dies every run with `synch primitive deleted while being used (ref=1)` at the `with_channel` line (secondary crash here is an AV in the leak dump walking the corrupted tracking list — the Windows flavor of the libc fatal). With `join()`: clean 3/3.

Verified the fence is complete, not just narrower: `NotifyAndRelease` drops `mRef` before decrementing `mRemaining` in one `mCompleteMutex` critical section, and `Wait()` reacquires that mutex — so `join()` returning happens-after the worker's release, and the nulled capture pointer means the lambda finalizer never touches the channel afterward. Same join-then-destroy idiom `with_wait_group` uses. Also confirmed the sibling consumer in `run_tool` (`for_each_clone`) is already fenced exactly as the doc addition says.

Thanks — good catch, and the skill addition reads as the right distinction rather than a blanket rule.

PR #3730, opened 2026-08-14
dasSpirv: ask the compiler what a shader const is worth under lint

olegus8:
A shader const spelled as arithmetic or as another const's name arrives at the
emitter unfolded under a lint compile, because that policy switches folding off.
`const_int_by_name` read `gv.init` directly, so it saw the expression rather than
the value and reported the const as not found; `emit_const` matched its own
shapes against the same unfolded node and failed a constant it could have
emitted.

Both now ask the compiler — `get_const_expr` (#3722) — instead of reading the
raw initializer:

- `const_int_by_name` asks once and matches the answer against the three
  constant shapes it already handled;
- `emit_const` asks **only where its own shapes matched nothing**, so a normal
  compile — handed an already-folded node — takes exactly the path it takes
  today, and a null expression stays the hard error it was.

`tests/spirv/test_const_fold_under_lint.das` compiles `_lint_fold_fixture.das`
under the lint policies and pins both halves: a shader whose group size comes
from a named const compiles and emits the right value, and the same for an array
literal whose elements are const arithmetic. The fixture keeps its consts
unfolded by construction, which is what makes the test meaningful.

Locally: `tests/spirv` is 374/374, lint and `dasfmt --verify` clean on the
changed files.

### DIFF
--- a/modules/dasSpirv/spirv/spirv_emit.das
+++ b/modules/dasSpirv/spirv/spirv_emit.das
@@ -843,19 +843,21 @@ def private const_int_by_name(name : string) : tuple<found : bool; value : int> 
     program_for_each_module(compiling_program()) $(mod) {
         for_each_global(mod) $(gv) {
             if (!got && gv.name == name && gv._type != null && gv._type.flags.constant && gv.init != null) {
-                let ci = gv.init ?as ExprConstInt
+                // under lint the initializer is unfolded, so what it is worth is the compiler's answer
+                let ce = get_const_expr(compiling_program(), gv.init)
+                let ci = ce ?as ExprConstInt
                 if (ci != null) {
                     result = ci.value
                     got = true
                 } else {
-                    let cu = gv.init ?as ExprConstUInt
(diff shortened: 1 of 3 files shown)


borisbat (approved):
Verified end to end: negative control (master's emitter + this test) fails and the PR passes it, so the test bites; the policy triple matches `utils/lint/main.das` verbatim, the real lint driver is clean on the shader-bearing fixture, and `tests/spirv` is 374/374 here. Also audited the fail-then-retry path for stale entries in `errors` — the only consumers that can hit it use throwaway probe arrays, so no spurious diagnostics.

One follow-up we'll land ourselves: the test invokes `compile_file` at runtime, and this directory marks that shape `options no_aot` (`test_conf_rejects.das`, `test_fail_closed.das`) so the nightly `test_aot` glob skips emission — the new test is missing the line and per-PR CI can't see it (the PR gate only AOT-compiles tests/language). One-liner, on us.

Thanks — good find, and the demand-fold consumption keeps the normal compile on its exact old path.

PR #3731, opened 2026-08-14
daspkg: copy a package without reading a file whole, and without its .git

olegus8:
`copy_dir_rec` is what `daspkg install` copies a package with, and it has three
problems:

1. **It recurses into `.git`.** Installing a module from a clone copies the whole
   history into `modules/<name>`. For the package that broke this for us that was
   125 MB of pack files against a few hundred KB of actual module.
2. **It reads every file whole**: `buf |> resize(sz); fread(fin, buf)`. That costs
   the file in memory, and past `max_unreserved_size` (64 MB) it does not degrade
   — it panics, and the install dies:

   ```
   EXCEPTION: array resize to 70000000 elements of 1 bytes each grows past
   max_unreserved_size (67108864 bytes)
     at daslib/builtin.das:74:8
   ```
3. **It sizes that buffer with `int(fs.size)`**, which truncates above 2 GB.

The copy now streams through a fixed 1 MB window and skips `.git`; sizing by the
file is gone, so the narrowing goes with it. Nothing else about the install
changes.

**How it surfaced.** #3724 brought the `max_unreserved_size` guard back to
`resize`, and the first sync that pulled it turned this latent bug into a hard
failure: our dasVulkan package is installed from a clone whose git pack is
125 MB, so every SDK install on both of our boxes died at the line above. The
guard is not at fault — it converted "quietly allocates 125 MB to copy a file
nobody wants" into a diagnostic, which is what a guard is for. Any package
carrying one large file reaches the same place, and a package that is a clone
gets there on its history alone.

**Verified against a tree carrying a 70 MB file and a `.git` beside it:**

| | pre-change | with this change |
|---|---|---|
| 70 MB file | the exception above | copied, byte-identical |
| 3 MB file (multi-window) | copied | copied, byte-identical |
| `.git/` | copied | skipped |

Gates: `utils/daspkg/utils.das` lints clean and passes `dasfmt --verify`.
`utils/daspkg/test_daspkg.das` is 218/219 here — identical on untouched master,
the one failure being `test_check_tool` 'detects cmake' on a box with no cmake.

### DIFF
--- a/utils/daspkg/utils.das
+++ b/utils/daspkg/utils.das
@@ -7,6 +7,8 @@ module utils shared public
 require daslib/fio
 require strings
 
+let COPY_WINDOW_BYTES = 1024 * 1024
+
 // ── Logging ──────────────────────────────────────────────────────────
 
 var log_file_path : string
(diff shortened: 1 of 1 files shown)


olegus8:
Correcting the body: I first wrote that nothing in daslang changed to cause this and that the guard predated it. That is wrong — `git log -S max_unreserved_size -- daslib/builtin.das` has exactly one commit, `f75f1ead6` from #3724, and it arrived in our tree with the sync that immediately preceded the failure. The bug in `copy_dir_rec` is older than the guard, but it was silent before it: reading a 125 MB file whole to copy it was always wrong, and #3724 is what made it say so. Body updated.

borisbat (approved):
Verified end to end on Windows: a probe tree with a 66.7 MB file (past the guard), a 3 MB + 7-byte multi-window file, an exactly-1 MB boundary file, an empty file, a nested dir, and a populated `.git` — all five content files SHA256-identical after the copy, `.git` skipped (the name check precedes `stat`, so a worktree-style `.git` *file* is skipped too), empty file created. The pre-change code on the same tree dies with exactly the exception in the description. `test_daspkg.das` is 219/219 on a box that has cmake, confirming your one failure as environmental.

A bonus the description undersells: the old code ignored the `fwrite` result, so a disk-full mid-copy produced a silent partial file — the new loop checks it.

Thanks — this also gives `is_package_content` as the natural seam for the other install-from-a-dev-checkout skips we have on the backlog.

PR #3732, opened 2026-08-14
jit: close four JIT-vs-AOT gaps (each(), generator range, array push, indirect dispatch)

aleksisch:
Four JIT codegen changes, each measured against AOT on the same binary, plus the probe suite that produced the numbers.

## Why

The JIT was slower than AOT on several everyday shapes for one structural reason: AOT gets `__forceinline` C++ templates where the JIT emits a call into the runtime. Each change below closes one of those, and each is a separate commit so it can be reverted on its own.

## What changed

**`jit: each(array/fixed array) for-loop sources walk inline`** — `for (x in each(arr))` hid an array behind a `Sequence`, paying the generic iterator protocol over a source this backend already walks natively. Unwrapped to the argument; the `each()` call is registered in `skipCall` and never evaluated, which also drops the per-loop heap iterator.

**`ast: a generator's for over a range keeps its counter in the capture frame`** — `replaceGeneratorFor` wrapped every source in `each()`, so `for (i in range(n))` inside a generator became a heap range iterator plus a virtual `first`/`next`/`close` per element **in all three lanes**. A range carries its own bound, so counter and bound live in the capture frame and `first`/`next` become the two lines `RangeIterator` would run. The counter is written through the same `void?` launder the C++ iterator used, so the loop variable stays `const` for the body — `cant_write_generator_for_variable.das` pins that (relaxing it silently turned `i = 5` into an infinite loop).

**`jit: array push-back inlines the fits-in-capacity path`** — AOT inlines `array_grow`, whose common case is a lock check, a capacity compare and a size store; only a real grow calls `array_reserve`. The JIT called `__builtin_array_push_back` on every push. The helper is kept for the grow and locked cases so growth policy and the locked-array throw stay in the runtime.

**`jit: inline the dispatcher on indirect calls to jitted functions`** — `jit_call_or_fastcall` only tests `SimFunction::jitFunction`, calls it and clears `Context::stopFlags`; that is now emitted in place, with the helper as the else-branch for interpreted/fastcall callees. **The two field offsets come from C++ accessors, not from constants in the emitter** — a hardcoded offset is a silent miscompile when the struct moves, a stale accessor is a link error.

Plus `ast: mark the generator-lowered launder nodes generated` (STYLE034 was reporting compiler-emitted IR as user source) and a correction to `skills/aot_testing.md`: `--use-aot` is required, there is no runtime auto-detection — a `test_aot` run without it silently interprets and still passes.

## Numbers

`benchmarks/core/jit_aot/primitives.das`, medians of 3, pinned core, both lanes on the same binary (ns/op):

| probe | AOT | JIT before | JIT after |
|---|---:|---:|---:|
| `each_iterator` | 1.4 | 2.3 | **0.5** |
| `each_iterator_setup/8` | 38.9 | 41.5 | **15.6** |
| `generator` (for over range) | 4.8 → **2.6** | 9.8 | **6.0** |
| `array_push` | 1.6 | 3.4 | **1.1** |

Indirect calls, interleaved same-session A/B with the emitter swapped between runs (JIT lane, medians of 3):

| probe | dispatcher | inlined |
|---|---:|---:|
| `class_method` | 3.00 | **1.60** (−47%) |
| `class_method_mono` | 2.90 | **1.60** (−45%) |
| `fnptr_invoke` | 2.80 | **1.60** (−43%) |
| `lambda_invoke` | 2.90 | **2.20** (−24%) |

Untouched probes (`static_call`, `try_recover`, `table_insert`, `build_string`) move ≤3%.

Across the 37-probe suite the JIT now wins or ties on most: `each_iterator` 0.33×, `static_call` 0.57×, `array_push` 0.69×, `table_int_get` 0.67×, `sort_default` 0.87×, `struct_clone` 0.84×, `build_string` 0.94×.

## Known remaining gaps (not addressed here)

- **Blocks** — `sort_block` (a comparator invoke per comparison) and `block_invoke` still take the generic path. Their capture base is not recoverable from a copied `Block` header; fixing it means storing the capture base inside `das::Block` (48 → 56 bytes), a real ABI change worth its own discussion.
- **`try_recover`** ~2.4× — two blocks built per `try` plus a runtime call, where AOT gets an inlined try/catch.
- **`hash_int`** ~2.1× — `_builtin_hash_int32` is `inline` in a header so AOT inlines wyhash. Matching it means reproducing wyhash bit-exactly in hand-written IR; since `hash()` feeds container hashing, a divergence between lanes would corrupt tables silently. Deliberately not attempted.
- Function-pointer/lambda/method calls still marshal arguments through a `vec4f` array where AOT passes them natively. Removing that needs either impl-symbol export (which risks the `static_call` win) or a per-module registration table.

## Testing

- `run_tests_jit` (full suite), `tests/language` under `-jit` (1606), `tests/jit_tests` (341)
- AOT lane via `test_aot_subset --use-aot` on `tests/language`
- New regression tests: `tests/jit_tests/each_inline.das` (9 arms incl. break/return unlock, mixed sources, nested loops), `tests/jit_tests/array_push_fast.das` (growth, reserved capacity, non-POD elements, locked-array throw), `tests/language/cant_write_generator_for_variable.das`, plus 3 arms in `tests/language/generators.das` covering every range flavour, two zipped ranges, and range zipped with a non-range iterator

Lint is untouched in this PR by request.

### DIFF
--- a/modules/dasLLVM/daslib/llvm_jit.das
+++ b/modules/dasLLVM/daslib/llvm_jit.das
@@ -3797,6 +3797,16 @@ class public LlvmJitVisitor : AstVisitor {
         return is_workhorse_table_key(tabT.firstType.baseType)
     }
 
+    def inline_each_source(expr : ExpressionPtr) : ExpressionPtr {
+        if (!(expr is ExprCall)) return expr
+        var cf = expr as ExprCall
+        if (cf.func == null || length(cf.arguments) != 1
+                || !starts_with(cf.func.name, "builtin`each`")) return expr
+        assume srcT = cf.arguments[0]._type
+        if (srcT == null || !(srcT.isGoodArrayType || srcT.baseType == Type.tFixedArray)) return expr
+        return cf.arguments[0]
+    }
+
     def build_table_lock_call(tab_ptr : LLVMOpaqueValue?; at : LineInfo) {
         var params = fixed_array(
             LLVMBuildPointerCast(g_builder, tab_ptr, types.LLVMVoidPtrType(), ""),
@@ -3886,7 +3896,7 @@ class public LlvmJitVisitor : AstVisitor {
             forBodyToExpr[expr.body as ExprBlock] = expr
         }
         for (ssrc in expr.sources) {
-            if (is_count_or_ucount(ssrc) || is_table_keys_values(ssrc)) {
+            if (is_count_or_ucount(ssrc) || is_table_keys_values(ssrc) || inline_each_source(ssrc) != ssrc) {
                 skipCall |> insert(ssrc as ExprCall)
             }
         }
@@ -3923,7 +3933,14 @@ class public LlvmJitVisitor : AstVisitor {
         LOOP_START:
         */
         var lblk = loop_stack |> back()
-        for (svar, ssrc in expr.iteratorVariables, expr.sources) {
+        for (ssource in expr.sources) {
+            let ssrc = inline_each_source(ssource)
+            if (ssrc != ssource) {
+                visit(ssrc, adapter)
+            }
+        }
+        for (svar, ssource in expr.iteratorVariables, expr.sources) {
+            let ssrc = inline_each_source(ssource)
             if (ssrc._type.isRange) {// range()->first()
                 var srange = getE(ssrc)
                 var sfrom = LLVMBuildExtractElement(g_builder, srange, types.ConstI32(0ul), "range_from_{svar.name}")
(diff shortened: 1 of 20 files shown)


modules/dasLLVM/daslib/llvm_jit.das
```diff
@@ -1745,6 +1745,13 @@ class public LlvmJitVisitor : AstVisitor {
                     params |> push(cmresPtr, 0)
                     var ccall = build_interop_call(expr.func, typ, func, expr.arguments, params, true, "")
                     setE(expr, cmresEval)
+                } elif (is_array_push_back(expr)) {
```
borisbat: there is very good chance this is not going to be the interop call which gets lowered. I'm pretty sure there are other ones. more generic solution - similar concept to how intrinsic are lowered - would be preferable

aleksisch: Now we have an easy way to add lowerings:
```
has_interop_lowering(expr)
lookup_interop_lowering(...)
```

PR #3862, opened 2026-08-25
WIP: ast-fuzz assertion detector, mutation kinds, and an ast_verify check for base-class nodes

aleksisch:
The AST fuzzer could not see the bugs it was built to find. Every sweep ran against a Release binary, where `DAS_NO_ASSERTIONS` compiles out all 318 internal assertions, so a run was only detectable when it segfaulted. This branch adds an assertions-enabled path, five mutation kinds, and a verifier check for the one defect class the work turned up.

The verifier now reports a base class used as a node. `ExprConst` and `ExprMakeLocal` are bases that a concrete node derives from, and neither overrides `visit`, so a walk reaching one asserts inside `Expression::visit` with no diagnostic. The walk it kills includes the verifier's own, so the scan reads the module's gc list instead of descending. This matters outside the fuzzer: a macro that builds an `ExprMakeLocal` crashes a Debug build today and gets a located error instead.

The fuzzer gained mutation kinds for function signatures and for arbitrarily deep type or statement nesting, and a `--mut-depth` knob. Four defects in the fuzzer's own instrumentation were fixed, each of which had been producing numbers that measured nothing: two concurrent sweeps shared one victim file, `--synth-skip` was inert for the main generation path, mutations aimed at a required module were never re-checked, and timeouts were judged against a baseline that excluded the macro module's own load cost.

No compiler bug was found. Roughly 11,000 source-expressible mutants across Release and Debug produced zero crashes and zero assertion failures, with the detector independently proven able to fire.

<details>
<summary>Ledger</summary>

### Validation

Full preflight and the AOT sweep were not run - this is a WIP branch and CI is the first full check. Run locally: rebase onto origin/master, lint and format over all changed `.das`, the ast-fuzz suite (19/19), and the make-pr sync and ast-verify gates. The Debug binary used for the assertion work was built in a separate worktree, since CMake pins output to `<src>/bin`.

The verifier check was negative-controlled: 60 real test files compiled under `--ast-verify` produced 0 false positives, and 26 of 31 generator repros that previously asserted now report a located error.

### Claims - stated, not tested

The remaining 5 repros reach the compiler's own `Program::visit` before any verifier entry runs. No daslib-side check can precede that walk; closing it would need a compiler-side pre-pass, which is not attempted here. A break would look like a Debug assert in `Expression::visit` with no `AST verify` line before it.

The ast-verify gate skipping `utils/internal/ast-fuzz/selftest/` is a pre-existing gap, not one this branch introduced: `cycle.das` and `alias.das` emit the same reports under the gate's own command.
</details>


aleksisch:
This PR turned out to be useless

PR #3921, opened 2026-09-01
interop: always hand a C++ bind "" for a null daslang string

aleksisch:
A daslang string is null when empty. Whether a bind taking `const char *` received that null depended on whether its translation unit included `ast/ast_typefactory_bind.h`, which is where the `cast_arg` specializations substituting `""` lived — fine while the call node was per bound function, broken once it became per signature: one game link has 27 objects defining the node for `void (*)(char const *)`, 10 with the substitution and 17 without, and object order picks the winner. `ImGui::PushID` got a `nullptr` and crashed in `ImHashStr`. Nothing diagnoses it — it is an ODR violation, and `ld.gold --detect-odr-violations` and `gcc -Wodr` are both silent.

So substitute unconditionally, in all three lanes:

- `cast_arg` specializations move beside the generic in `simulate/interop.h`; the AOT and JIT emitters stop asking whether the callee opted in, which retires `needStringCast`
- the argument null-checks that can no longer fire go: `module_builtin_string` trades 44 `stringLengthSafe` for `stringLength` and drops 22 guards. What stays is what the substitution does not reach — a string taken by reference, non-bind helpers, and guards on lengths and `FILE *`
- `ast_typefactory_bind.h` is left empty, so `typeFactory` for unbound callbacks moves next to `typeFactory<TT *>` in `ast_typedecl.h` (that order matters: above it, GCC rejects the partial specialization), `cast_arg<das::string>` moves to `interop.h`, and the header, its 102 includes and the binder line emitting it all go

Cost, from the second commit's benchmark — two binds, four `const char *` against four `void *`, same body and node: **10.3 vs 10.1 ns/op** interpreted, 0 allocations either way.

Two notes for review:

1. Removing the flag bit leaves a hole in `MoreFunctionFlags`, which rtti exposes and the AST module cache serializes — may want a cache version bump.
2. Empty results still come back as null (`allocateString` cannot allocate an empty string), but a null crossing into a bind and back — as through `_temp_string_result`, wrapped around every `[temp_string_result]` builtin — returns as `""`. So `is_null_str` in the byte-view pins measures length now, and one assertion asking whether a NUL-holding run was allocated at all is gone: das can no longer tell. Covering that again needs a C++-side test calling the builtin directly.

Out-of-tree generated bindings include the deleted header, so this needs borisbat/dasImguiImplot#30 and borisbat/dasImguiNodeEditor#46 first — `fatman` and `wasmboy` clone their default branches.


aleksisch:
Closing - opened without the author's go-ahead. Will re-file when asked.

include/daScript/simulate/simulate_nodes.h
```diff
@@ -532,6 +532,18 @@ namespace das {
 #pragma diag_suppress 826
 #endif
 
+    struct SimNode_StringArgNotNull : SimNode {
+        DAS_PTR_NODE;
+        SimNode_StringArgNotNull ( const LineInfo & a, SimNode * se ) : SimNode(a), subexpr(se) {}
+        __forceinline char * compute ( Context & context ) {
+            DAS_PROFILE_NODE
+            char * res = subexpr->evalPtr(context);
```
borisbat: this is going to be slower than original implementation, which uses cast<>
cast gets inlined on C++ side, this creates interpreter node - which is a lot slower
even blanket cast_arg<char *> would be faster

borisbat (dismissed):
way too slow.

include/daScript/simulate/simulate_nodes.h
aleksisch: Now we simply wrap it everywhere on das -> cpp path

include/daScript/simulate/interop.h
```diff
@@ -12,6 +12,30 @@ namespace das
...
+        static __forceinline char * to ( Context & ctx, SimNode * node ) {
+            char * res = node->evalPtr(ctx);
+            return res ? res : ((char *)"");
+        }
+    };
+
+    template <>
+    struct cast_arg<das::string> {
+        static __forceinline string to ( Context & ctx, SimNode * node ) {
+            auto res = (das::string *) node->evalPtr(ctx);
```
borisbat: this is a bit tricky. is this properly supported in AOT / JIT?
i.e. this allows void foo ( string bar ) on C++ side to be bound.

borisbat (approved):

include/daScript/simulate/interop.h
aleksisch: I agree. It was here earlier, so I just kept it. Let's remove?

aleksisch: Removed

PR #4049, opened 2026-09-16
infer: three fuzzer crashes from #4044

### LINKED ISSUE #4044: Fuzzer crashes (3)
Grammar-based fuzzer findings, continuing from #3959. Tracking: #2136.

Built from `781b5e90e`, stock `RelWithDebInfo`. All three reproduce with plain `daslang <file>` — no `-aot`, no `-jit`, no `options`.

### 1. A class deriving from a class whose own parent was rejected crashes in `makeClassRtti`

```das
struct A;
class B : A;
class C : B;
```

```
$ daslang t.das
CRASH: SIGSEGV (Segmentation fault) (signal 11) at address 0x28

#0  das::makeClassRtti(das::Structure*)
#1  das::ast_structureDeclaration(...)
#2  das2_yyparse(void*)
#3  das::parseDaScriptEx(...)
```

`class B : A` is rejected as it should be (`error[30134]: class can only derive from a class`), and that rejection returns from `ast_structureDeclaration` before `makeClassRtti(B)` runs — so `B` never gets a `__rtti` field. `class C : B` is then accepted, and `makeClassRtti(C)` writes through `findField("__rtti")`, which is null.

All three lines are required: `struct A; class B : A;` on its own gives the clean error.

### 2. An operator the typer never resolved leaves a null `func`, crashes in `buildAccessFlags`

```das
def main {
    1 &&= typeinfo safe_has_field<var48>(typeinfo is_ref(1))
}
```

```
$ daslang t.das
CRASH: SIGSEGV (Segmentation fault) (signal 11) at address 0x1c6

#0  das::TrackFieldAndAtFlags::getSideEffects(das::Function* const&)
#1  das::TrackFieldAndAtFlags::preVisit(das::ExprOp2*)
#2  das::ExprOp2::visit(das::Visitor&)
...
#7  das::Program::buildAccessFlags(das::TextWriter&)
```

`preVisit(ExprOp2*)` and `preVisit(ExprOp1*)` in `ast_unused.cpp` call `getSideEffects(expr->func)` unconditionally; `preVisit(ExprOp3*)` and `preVisit(ExprCall*)` right below them already guard on a null `expr->func`.

Three ingredients, all required:
- `&&=` — without it the same expression compiles and the program exits 0
- an undefined type as the type argument — `safe_has_field<int>` does not crash
- a nested `typeinfo` as the value argument — a literal does not crash

### 3. A fixed-array dimension naming the variable it declares eats all memory

```das
let zz : int[zz][zz]

def main {
    var cb : array<int>
    cb := cb
}
```

```
$ daslang t.das
(no output; killed by the OOM killer)

$ dmesg | tail -1
Out of memory: Killed process 2841238 (daslang_static) total-vm:39511540kB, anon-rss:26836940kB
```

26.8 GB RSS from a three-line program. Under `( ulimit -v 4194304; ... )` it dies as a SIGSEGV at ~2.7 GB in 15s, with this repeating:

```
#3  das::ExprVar::clone(das::Expression*) const
#4  das::TypeDecl::TypeDecl(das::TypeDecl const&)
#5  das::TypeDecl::TypeDecl(das::TypeDecl const&)
#6  das::ExprVar::clone(das::Expression*) const
...
#49 das::InferTypes::visit(das::ExprVar*)
#52 das::Program::visitGlobalVariable(das::Visitor&, das::Variable* const&)
#55 das::inferTypesDirty(das::Program*, das::TextWriter&, bool)
```

`InferTypes::visit(ExprVar*)` gives the dimension's `ExprVar` the variable's own type, which contains that same `ExprVar` — so every later deep copy of the type walks the cycle. The existing `globalVar` self-reference check (`global variable '...' can't be initialized with itself`) does not fire here, because `globalVar` is only set for the initializer, and the type is visited before it.

Two ingredients, both required:
- **two** dimensions. `let zz : int[zz]` errors out in 0.42s
- an `array<>` clone (`:=`) somewhere in the program. With `<-`, or with no clone at all, it errors out in 0.42s

---

The same fuzzing run also produced 16 inputs that hard-crash on deeply nested types (`TypeDecl`'s copy constructor and `Expr*::clone` recursing until the stack is gone) — that is #3144, not re-reported here.

aleksisch:
Fixes #4044 — all three crashes, one fixture each, plus the root cause behind the second one.

## 1. `makeClassRtti` null deref

```das
struct A;
class B : A;
class C : B;
```

`class B : A` is rejected as it should be, and that rejection returns from `ast_structureDeclaration` **before** B gets its generated `__rtti`. `class C : B` is then accepted and wrote through a `findField("__rtti")` that answers null.

`makeClassRtti` and `makeClassFinalize` now let the parent lookup pick the branch: a parent that never got the field means the class declares its own. `makeClassFinalize` had the identical defect one call later and would have been the next crash.

→ `error[30134]: class can only derive from a class`, one error, no crash.

## 2. Null `ExprOp2::func` in `buildAccessFlags`

The reported crash is a missing guard, but it was hiding a real inference bug.

**Root cause.** Four `safe_*` `typeinfo` traits returned a replacement node **without `reportAstChanged()`** — `safe_has_field`, `struct_safe_has_annotation`, and both arms of `struct_safe_has_annotation_argument`. Every sibling branch signals; these four did not. The replacement arrives untyped, the enclosing operator takes its `failed to infer` early-out, and with no signal that pass is the last one — so `ExprOp2::func` stays null with no diagnostic.

It broke valid programs, not just fuzzer input:

```das
var b = true
b &&= typeinfo safe_has_field<X>(typeinfo is_ref(1))    // did not compile
```

**Guard.** `preVisit(ExprOp1*)` / `preVisit(ExprOp2*)` in the access-flag pass now check `func` the way `ExprOp3` and `ExprCall` already did. `buildAccessFlags` runs before the side-effect pass that owns the `50613` diagnostic, so without the guard any unsignalled rewrite crashes before it can be reported.

→ the fuzzer input now gives `error[30341]: no matching functions or generics: _::&&=(int const, bool const)`.

A fifth site, `is_iterable`, looked like the same defect but already signals three lines above its return — left alone.

## 3. Self-referential fixed-array dimension (OOM, 26.8 GB)

```das
let zz : int[zz][zz]
```

The dimension expression lives *inside* the type it dimensions, so giving the dimension's `ExprVar` the variable's own type gives it a type containing that same expression. With two dimensions each infer pass squares the previous pass's copy — 4^50.

`InferTypes::visit(ExprVar*)` no longer adopts a global's type while its dimensions are unresolved (`isExprType()`).

This also covers two shapes the report does not name, both of which OOM'd on master and which the `globalVar` self-reference guard suggested in the issue would not have caught:

```das
let aa : int[bb][bb]                  // mutual
let bb : int[aa][aa]

let aa : int[length(bb)][length(bb)]  // through a call
let bb : int[length(aa)][length(aa)]
```

→ `error[30109]: array dimension must be constant`, twice, in 0.4s.

## Tests

| fixture | pins |
|---|---|
| `failed_class_parent_rejected.das` | 1, `expect 30134` |
| `failed_op2_int_lhs_typeinfo_rhs.das` | 2, the issue's exact input, `expect 30341` |
| `failed_global_dim_self_reference.das` | 3, `expect 30109:2` |
| `typeinfo_safe_has_field_reinfer.das` + 3 siblings | one `typeinfo` branch each |
| `failed_macro_added_op2_must_infer.das` | op1/op2 with a null `func` reach the `50613` diagnostic |

**One shape per file is load-bearing** for the four `typeinfo` fixtures: the trait has to fold in a pass where nothing else changes, and a second shape in the same file signals for the first. My first attempt was a six-arm dastest file and it passed without the fix — a tautology. Negative control: rebuilt with the four `reportAstChanged()` calls reverted, all four files fail.

Two things that are coverage rather than controls, stated plainly:

- the `ast_unused` guard is not reachable by a test once the root cause is fixed — it is consistency with `ExprOp3`/`ExprCall`, as the issue asked for
- `failed_macro_added_op2_must_infer.das` does not fail without the guard (`buildAccessFlags` runs before `patchAnnotations`); it covers two previously untested `reportNullFunc` sites

## Verification

- `tests/language`: 1759/1759
- full `--test tests`: 14530 tests, 14514 passed, 0 failed, 0 errors, 16 skipped
- formatter `--verify` clean over `tests/language` (410 files); lint clean on the new files; `src/ast/REVIEW.das` gate OK
- `test_aot_subset` builds with the new fixtures; the `failed_*` files are excluded by the existing filter at `tests/aot/CMakeLists.txt:334`, so no CMake change


src/ast/ast_infer_type.cpp
```diff
@@ -2716,6 +2716,7 @@ namespace das {
                     return new ExprConstBool(expr->at, ft != nullptr);
                 } else {
                     if (expr->trait == "safe_has_field") {
```
borisbat, line 2718: how so? why is this an AST change??

src/ast/ast_infer_type.cpp
```diff
@@ -2730,6 +2731,7 @@ namespace das {
                     return new ExprConstBool(expr->at, it != ann.end());
                 } else {
                     if (expr->trait == "struct_safe_has_annotation") {
+                        reportAstChanged();
                         return new ExprConstBool(expr->at, false);
```
borisbat, line 2735: same

src/ast/ast_infer_type.cpp
```diff
@@ -2742,6 +2744,7 @@ namespace das {
                     auto it = find_if(ann.begin(), ann.end(), [&](const AnnotationDeclarationPtr &pa) { return pa->annotation->name == expr->subtrait; });
                     if (it == ann.end()) {
                         if (expr->trait == "struct_safe_has_annotation_argument") {
+                            reportAstChanged();
                             return new ExprConstBool(expr->at, false);
```
borisbat, line 2748: same again. why is this an ast change?

borisbat (commented):
mainly why is type-trait an ast-change?

PR #4053, opened 2026-09-17
bench-stand: the jit lane stops rebuilding the JIT emitter once per benchmark

aleksisch:
The jit lane cost three times the interp lane and almost none of it was measurement. Every child process rebuilt the JIT emitter - `llvm_jit.das` and the ~52 files of dasLLVM around it - before it could compile a benchmark, and the bench stand spawns one process per file.

The DLL cache is keyed per script, so every file pays the emitter build again. Across 68 benchmarks that is ~53 minutes a night spent compiling the emitter rather than measuring anything.

A standalone dastest carries the emitter as compiled code. What it did not carry was the llvm/dasbind machinery a JIT compile needs, and `jit_enabled()` is true inside any exe whether the machinery is there or not. Having the exe detect its own capability does not work: `has_module("llvm")` is false even under a working `-jit`, because the llvm module loads on the first JIT compile - nothing distinguishes the two before one is attempted. So the host declares it: `--jit-register-all-modules` at build time, `--jit` at run time. `jit` and `lint` already take that build flag for the same reason; dastest joins them.

The guard is a pure function of the suite context, `tests_jit_enabled(ctx)`, not module state behind a setter.

One benchmark file, same arms:

| | `daslang -jit` | native dastest |
|---|---|---|
| wall time | 123.75 s | **7.76 s** |
| `array_rw/100000` | 0.6 ns/op | 0.9 ns/op |
| `array_sum/100000` | 0.1 ns/op | 0.1 ns/op |

Still jit numbers - the interpreter reads 6.2 and 4.8 on those arms, which is also what the exe reports when `--jit` is absent, so the flag is a real switch rather than always-on.

The lane takes that host when the tree has one and falls back to `daslang -jit` when it does not, so a tree that never built it still runs. A benchmark requiring a dynamically loaded module - pugixml, sqlite, terminal - cannot run from an exe that links none of them and takes the same fallback; that is five files of seventy, found by the child's own error rather than by a list that would drift. The interp lane keeps the daslang host: its startup was ~2.5 s, so the exe buys nothing there and the fallback retries cost more than they save.

**An aot lane.** daslang runs three ways and the stand measured two, leaving the tier a shipped game executes unplotted. The lane spawns `test_aot_subset` under `-use-aot`. What is compiled into that host is what it can measure - `tests/aot/CMakeLists.txt` registers `benchmarks/core/jit_aot` and nothing else under `benchmarks/` - so the other groups are excluded per group with their reason. A file outside the set is skipped in 0 s: `--use-aot` on a file with no stubs would either burn 76 s failing or quietly file interpreter numbers under an AOT label.

**The nightly clean kept deleting the module binaries.** It names where build output lives - `build`, `bin`, `lib`, `.jitted_scripts` - but a module's binary is written beside its sources as `modules/<name>/*.shared_module`. The build rewrites them a minute later so nothing notices; under `BENCH_STAND_BUILD=skip`, the documented local dry run, nothing does and the tool cannot load a module at all.

## What the lane's own CI runs turned up

Standing the lane up against the full matrix surfaced defects in the tiers it measures and in the lanes around it. They are separate commits.

**JIT correctness.** `try`/`recover` was lowered during inference, which mutated the AST the semantic hash is taken over, so an AOT object and its consumer disagreed - `error[50101]: AOT link failed`. Lowering moved to the backend, leaving infer's AST untouched. An emitted AOT object is now keyed by the tree its consumer rebuilds rather than by the emitting tree, and the emitter's finalizer shuts LLVM down so a `-jit` run exits clean instead of at a destructor.

**The JIT DLL cache corrupted itself under concurrency.** Objects were written in place and collected by age, so one run read a partially written artifact another run was still emitting, and the collector deleted work in flight. Artifacts now publish by `rename(2)` and the collector only removes names the run itself published - nothing else is its to delete. Verified with twelve concurrent copies.

**dasHV fixtures fought over ports.** `run_all` drives the interpreter, jit and aot lanes over `tests/` at once, so up to four copies of one dasHV test run together, and every fixture bound a fixed port. The losers reported `bind: Address already in use` and then served from a server that never started - one run took SIGSEGV at 0x120 in the following tick. A server now asks for port 0 and reports the port it got, the way Go's `net.Listen(":0")`, Rust's `TcpListener::bind("…:0")` and Java's `new ServerSocket(0)` do. libHV listens only above zero and its `listenfd` is an input it adopts rather than an output, so `start()` learns a free port from a socket of its own, releases it, and binds that number; `bound_port` reads it back. The fixture publishes the port over a one-capacity channel whose pop blocks until the value lands - one primitive where a wait group plus an atomic used to carry the wait and the number separately.

**The live-host wait was asking the wrong question.** It built an HTTP client per poll; under load one poll took 64 s, so a host ready in 1.6 s was seen at 47 s, 96 s or not at all. It now asks the port whether anything is listening.

## Lanes that were doing work twice

- `run_doc` invoked das2rst twice, racing itself; the callers outside cmake run the target instead, and preflight checks the digest.
- The shipped module digest is generated to the build dir and checked against the tree, so a doc build no longer rewrites a tracked file as a side effect.
- The standalone sweeps cache on what they read - 533 s to 1 s on an unchanged tree.
- `stage_doc_assets` refetched all 68 tutorial MP4s on every doc build, 29 s of network for a set that has changed once since July. A staged file's mtime is when it was downloaded, so only an asset uploaded since reads newer; one metadata call answers it for the whole set, and an unchanged tree now costs 1.3 s and no download.
- A lane's leaves wait on the modules the lane loads, so a lane no longer reads a `.shared_module` another lane is relinking.
- The mingw self-binder reads the clang headers from the subsystem prefix, and the bundle smoke no longer restores a stranded `lib/` at startup.
- `ADD_EXAMPLE_RUN` printed a line per call site at configure time; the list it builds is what matters.

## Testing

`utils/benchctl/tests` and the lanes were exercised directly: the jit lane over `benchmarks/micro` went 618 s to 390 s with three of its eight files on the pugixml fallback, and per file the fast path ran 63.4 -> 4.5 s, 50.3 -> 3.7 s, 60.5 -> 6.7 s, 119.1 -> 25.5 s, 130.2 -> 30.2 s. The aot lane runs `core/jit_aot/primitives.das` green with 37 arms and skips a non-registered file in 0.0 s with its reason.

The dasHV suites pass singly and under twelve concurrent copies, with no fixture directory left behind. The DLL cache passes twelve concurrent copies with no panic. `check_docs_fresh` reports the generated documentation fresh, and `check_lint_changed` reports 0 issues over this branch's files.

A full local `run_all` has not completed end to end - it was interrupted several times while this branch was being rebased and reworked. CI is the check that matters.

#nightly


borisbat (approved):
looks fine

aleksisch:
We merge it even though it doesn't fix entirely. There will be follow up PR.

PR #4065, opened 2026-09-18
ci: every CI check becomes a cmake target, and the workflows become setup plus calls

aleksisch:
Draft.

## What this does

Every CI check becomes a cmake target, defined in the directory it checks. A lane's step and a developer's command are now the same line.

Before, the workflows carried ~45 raw `dastest` invocations, a dozen shell and python steps, and four bash `case` ladders in `build.yml` that no local mirror could reproduce. `skills/internal/preflight.md` now names the target in every mirror row, including the dozen rows that previously had no local mirror at all.

## The commits

1. **`das-fmt: a file that disappears between the scan and the open is a skip, not a failure`** — a pre-existing bug on master, unrelated to the rest but hit constantly once lanes run in parallel. `das-fmt` collects the whole file list in one walk, then spends the run opening entries from it; anything deleted in between was reported as `Unable to open`, and on the comment-stripping path it incremented `strip_failures`, which `main` turns into `exit(1)`. Both paths now warn and skip. A path named explicitly on the command line keeps its error and now reports at collection time, so a mistyped `--path` fails immediately.
2. **`cmake: the install gate's regexes escape their dots`** — three `\.` in quoted cmake arguments were being silently stripped to `.` under CMP0010.
3. **`nano: PerfectHash::build passes das::max two arguments of the same type`** — another pre-existing master bug. `uint32_t` is `unsigned int` on the hosted targets and `unsigned long` on `arm-none-eabi`, so `das::max(1u, (n + 4) / 5)` deduces one template argument natively and two different ones when nano cross-compiles for cortex-m4, where it stops compiling. Only `check_nano_arm` reaches it, and that step ran nightly-only before this branch.
4. **`ci: every CI check becomes a cmake target`** — the refactor itself.

## Placement

Placement follows ownership: `utils/` owns the tool suites, the lint and ast-verify gates and the REVIEW.das gates; `tests/` the sweeps; `tests-cpp/` the ctest wrappers; `examples/games/` the wiring gate, boulder-dash and the sequence smoke; `doc/` the rst generators, asset staging and doc-verify; `site-dasllama/`, `benchmarks/`, `nano/`, `modules/dasGlfw` and three other modules their own.

Where a check needs logic a target cannot express, the script text lives in that directory's `CMakeLists.txt` and configure emits it, so no directory holds a loose `.cmake` beside its list file. `cmake/das_lane.cmake` is the one shared module, included from the root because `modules/` is processed long before `utils/`.

`ci/CMakeLists.txt` keeps only what is genuinely ci — the tests of its own python scripts and the bundle smokes — plus the lane targets: `run_build`, `run_build_nightly`, `run_extended_*`, `run_doc`, `run_site`, `run_nightly` and `run_all`. Each skips a target this configure does not define, and says which at configure time.

## The workflows

**`build.yml`** loses its four `case` ladders. The configure is one command for every cell:

```bash
cmake --no-warn-unused-cli -B build -G "${{ matrix.cmake_generator }}" \
  -DCMAKE_BUILD_TYPE:STRING=${{ matrix.cmake_preset }} \
  -DDAS_LLVM_DISABLED=${{ env.das_llvm_disabled }} \
  -DDAS_USE_SANITIZER=${{ matrix.sanitizers }} \
  -DDAS_FAST_MATH=${{ matrix.fast_math }} \
  ${{ matrix.cmake_extra }}
```

`matrix.cmake_extra` carries the only genuinely per-cell part: the compiler, the arm module flags, the darwin architecture, and the windows launcher plus vcpkg toolchain. Three of the five old branches differed in values only.

`run_tests_sweeps` decides its own sweeps from settings cmake already has — no JIT where dasLLVM is off or on Debug, no interpreter sweep under a sanitizer. The retry stays a `||` in the workflow, where bash already has one.

**`extended_checks.yml`** runs one target per job, `run_extended_${{ matrix.role }}`. The core/modules split had taxed every check with a `matrix.role != '<other>'` condition; membership moved into `ci/CMakeLists.txt`. 47 step conditions become 14, and all 14 are provisioning. The file goes from 700 lines to 288. The python packages it needs become `ci/requirements.txt`, the way `doc/` already keeps its own.

## Lanes are dependency aggregates

Every lane target is a plain `add_dependencies` aggregate. No lane re-enters the build directory with a nested `cmake --build` — two ninjas in one build dir take no lock on it and will relink the same outputs on top of each other, which surfaces as `file too short` on a shared library at a completely unrelated call site.

## check_bundle_smoke is deliberately in no lane

`ci/smoke_test_bundle.sh` renames the build tree's `lib/` to `lib.smokehidden/` for the duration of its run. It has to: every exe that `daslang -exe` produces carries `-Wl,-rpath,<source>/lib`, so on the machine that built the bundle a build-tree rpath resolves and the launch check would pass without the bundle-relative entry ever being exercised.

That makes the check **globally exclusive** — while it runs, every link and every `daslang` launch anywhere in the tree fails. That was harmless when CI ran steps one at a time; it is not harmless in a parallel lane. So `check_bundle_smoke` is a member of no lane group, and `run_all` does not cover it. `build.yml` already invokes it as its own step, so CI loses nothing. Locally it is two commands:

```
cmake --build build --target run_all -j 16 -- -k 0
cmake --build build --target check_bundle_smoke
```

The script's `trap ... EXIT` restores `lib/` on normal exit and on Ctrl-C, but not on `SIGKILL` or a hard ninja abort — which strands the tree in a state whose symptoms point everywhere except the cause. It now also restores a leftover `lib.smokehidden/` at startup, so a killed run self-heals on the next one.

Making the smoke safe to run concurrently is a separate change, and would mean either isolating the rpath (a mount namespace, or a build tree of its own) or replacing the hide-and-launch check with a static `RPATH`/`RUNPATH` assertion on the installed binaries.

## Guardrails

`ci/test_ci_matrix.py` is unchanged from master except for one test. `test_nightly_only_steps_are_exactly_the_pinned_set` read per-step `if:` conditions that the refactor removes; it now pins the six nightly-only checks by **target name**, read from `DAS_EXTENDED_NIGHTLY` in `ci/CMakeLists.txt`. A check that joins no group list is a red test rather than one that silently stopped running. `ci/REVIEW.das` also reads `ci/CMakeLists.txt`, so a `ci/test_*.py` wired only to a target still counts as wired.

## Verified

- The new sweep policy matches the old shell decision for all 15 matrix cells, both decisions each, zero mismatches.
- `check_format_exe` over the whole tree: 5300 files, clean.
- `das-fmt`'s two new branches probed directly: a listed-then-removed file warns and exits 0; a missing `--path` errors and exits 1. Format-verify and lint clean on the edited file.
- Every workflow parses as YAML.
- `check_nano_arm` cross-compiles clean for cortex-m4 after the `das::max` fix, and the same translation unit still passes a native syntax check with the build's own flags.
- A negative control caught a real bug during development: the per-file AST verify had applied the `ast-fuzz/selftest/` skip in both modes, where the original applies it only to the changed set.

## Not verified

No full build or lane run since `run_tests_cpp_small`/`run_tests_cpp_big` were folded into `test-small`/`test-big`. That fold deleted two targets and repointed five `build.yml` call sites.

## Not done

- Three workflows still build the module flags in shell: `release.yml` twice, `dasllama_server_release.yml`, `nightly_daspkg_index.yml`.
- `--config` is inert on every Ninja lane and could be deleted from 34 places; only `msvc.yml`, the one Visual Studio lane, needs it.
- The mingw and clang-cl nightly jobs still run their sweeps as hand-rolled steps. Folding them in is not behaviour-preserving, so it wants its own change.
- Setup duplication remains: the Defender block appears 10 times, sccache setup 11, the apt list 17. Composite actions are the fix, separate from this.


aleksisch:
Still not always works:
```
cmake --target run_all
```
But we never run it in CI concurrently, so it's fine. Follow-up will be presented to fix all the inconsistencies.

